PR #1, opened 2026-04-17
Migrate vllm-spyre-next plugin to standalone spyre-inference repo

tdoublep:
## Summary
- Copies the `vllm-spyre-next` plugin from the `sendnn-inference` monorepo into this standalone repository
- The only change from the source is updating `setuptools_scm` root from `..` to `.` for standalone repo use
- All original contributors are included as co-authors

## Test plan
- [ ] Verify `pip install -e .` succeeds and `import vllm_spyre_next` works
- [ ] Run `pytest tests/` to confirm test collection works
- [ ] Verify `uv lock` succeeds (or note if it requires CI environment)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


joerunde:
@tdoublep there are also docs for the new plugin under docs/vllm_spyre_next if you wanna grab those too

PR #3, opened 2026-04-21
renaming vllm-spyre-next -> spyre-inference

yannicks1:
closing [938](https://github.com/torch-spyre/spyre-inference/issues/9)


joerunde:
Couple bits in pyproject.toml:

```
pyproject.toml:description = "Next iteration of vllm-spyre on the torch-spyre stack"
pyproject.toml:# Only match spyre-next-vX.Y.Z tags (with optional SemVer pre-release suffix), ignoring v* tags used by vllm-spyre.
```

then the docs are still nested under `docs/vllm-spyre-next`. We should be able to just remove the `vllm-spyre-next` nesting there

pyproject.toml
```diff
@@ -23,29 +23,29 @@ requires-python = ">=3.11"
...
+version_file = "spyre_inference/_version.py"
 # Version strings are formatted as `{version}+{local}`, e.g. "0.1.dev1+gb1a7032"
 # Local versioning is incompatible with pypi, so we disable it by default.
 local_scheme = "no-local-version"
 root = ".."
-# Only match spyre-next-vX.Y.Z tags (with optional SemVer pre-release suffix), ignoring v* tags used by vllm-spyre.
+# Only match spyre-next-vX.Y.Z tags (with optional SemVer pre-release suffix).
 # SemVer pre-release format: hyphen followed by dot-separated identifiers (alphanumeric + hyphens)
 tag_regex = "^spyre-next-v(?P<version>[0-9]+\\.[0-9]+\\.[0-9]+(?:-[0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*)?)$"
 git_describe_command = "git describe --dirty --tags --long --match 'spyre-next-v*'"
```
sducouedic: Not sure exactly how this works, but shouldn't we change to `spyre-inference` as well? 

```suggestion
# Only match spyre-inference-vX.Y.Z tags (with optional SemVer pre-release suffix).
# SemVer pre-release format: hyphen followed by dot-separated identifiers (alphanumeric + hyphens)
tag_regex = "^spyre-inference-v(?P<version>[0-9]+\\.[0-9]+\\.[0-9]+(?:-[0-9A-Za-z-]+(?:\\.[0-9A-Za-z-]+)*)?)$"
git_describe_command = "git describe --dirty --tags --long --match 'spyre-inference-v*'"
```

sducouedic (commented):
See my comment in pyproject, beside of that I think it looks good

sducouedic (approved):
thanks Yannick, LGTM

pyproject.toml
```diff
@@ -23,31 +23,31 @@ requires-python = ">=3.11"
...
 [tool.setuptools_scm]
 # Version files are optional, this just marginally speeds up version inspection
 # at runtime
-version_file = "vllm_spyre_next/_version.py"
+version_file = "spyre_inference/_version.py"
 # Version strings are formatted as `{version}+{local}`, e.g. "0.1.dev1+gb1a7032"
 # Local versioning is incompatible with pypi, so we disable it by default.
 local_scheme = "no-local-version"
-# Only match spyre-next-vX.Y.Z tags (with optional SemVer pre-release suffix), ignoring v* tags used by vllm-spyre.
+# Only match spyre-inference-vX.Y.Z tags (with optional SemVer pre-release suffix).
```
joerunde: We shouldn't need to do this at all- there's only one project in the repo now so the tags can just be the standard `vX.Y.Z`.

Docs for this are [here](https://setuptools-scm.readthedocs.io/en/latest/config/), and the default should be fine. Let's remove both `tag_regex` and `git_describe_command`

(We can fix in a followup though, ljgtm)

joerunde (approved):
very nice 😎

PR #132, opened 2026-04-24
[Spyre-Next] On-device KV Cache

jvlunteren:
## Description

vLLM’s paged KV cache relies on advanced tensor indexing to scatter newly generated tokens into arbitrary physical block locations and later gather them back into logical order for attention. Spyre does not yet support these operations (advanced indexing fails to execute on-device).

As a temporary workaround, I replaced the indexing-based logic with two primitives that Spyre does support: element-wise arithmetic and matrix multiplication. These primitives are used to implement the required scatter and gather functionality. This approach enables an initial implementation of an attention kernel in which the KV cache resides entirely on Spyre, avoiding CPU involvement. This initial implementation is intended for functional correctness rather than performance optimization.

Due to current Spyre limitations, the implementation operates under several constraints:
- batch size: `max_num_seqs=1`
- `num_gpu_blocks_override` must be set (e.g., to 64)
- `max_model_len <= num_gpu_blocks_override * block_size`
(e.g., with `block_size=16` and `num_gpu_blocks_override=64`, `max_model_len=1024` tokens)

The on-device KV cache implementation lives in `spyre_attn_exp.py` and can be activated by setting `SPYRE_ATTN_IMPL=exp`. The default backend remains unchanged when the variable is unset.

A `--num_gpu_blocks_override` CLI argument was added to `torch_spyre_inference.py` (defaults to None, letting vLLM auto-calculate) to allow explicit control over the KV cache block count when needed for testing the on-device backend.

## Test Plan

Verified correct output with the default backend:
`python3 torch_spyre_inference.py --model ibm-granite/granite-3.3-8b-instruct --max_model_len 1024 --max_num_seqs 1 --num_gpu_blocks_override 64 --num-prompts 1 --attention_backend CUSTOM --custom_ops +RMSNorm +MergedColumnParallelLinear +RowParallelLinear +VocabParallelEmbedding`

Re-ran with `SPYRE_ATTN_IMPL=exp` and confirmed matching results.

Output:
```Processed prompts: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [13:33<00:00, 813.68s/it, est. speed input: 0.01 toks/s, output: 0.02 toks/s]
Time elapsed for 20 tokens is 814.03 sec
===============
CompletionOutput(index=0, text="\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,", token_ids=[203, 203, 36671, 1182, 2575, 6866, 46654, 2305, 6564, 21909, 11258, 30, 6564, 17265, 11258, 30, 3859, 30, 18919, 30], routed_experts=None, cumulative_logprob=None, logprobs=None, finish_reason=length, stop_reason=None)
===============

Prompt:
 'What are IBMs main businesses?'

Generated text:
 "\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,"

-----------------------------------
(EngineCore pid=58079) INFO 04-24 12:39:55 [core.py:1210] Shutdown initiated (timeout=0)
(EngineCore pid=58079) INFO 04-24 12:39:55 [core.py:1233] Shutdown complete
```

## Checklist

- [x] I have read the [contributing guidelines](https://docs.vllm.ai/projects/spyre/en/latest/contributing)
- [x] I have updated the documentation (if applicable)


spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,591 @@
...
+        )
+
+        self._k_vals_buf = torch.zeros(
+            self.num_kv_heads, self._aligned_num_physical_blocks, self._block_width,
+            dtype=self._target_dtype,
+        )
+        self._v_vals_buf = torch.zeros(
+            self.num_kv_heads, self._aligned_num_physical_blocks, self._block_width,
+            dtype=self._target_dtype,
+        )
```
bohnstingl: IIUC instead of a single KV cache tensor, we create here multiple smaller, but dedicated ones, right? In the current attention implementation and also in other upstream implementations, the format is typically `(num_blocks, 2, block_size, num_kv_heads, head_size)` or (`2, num_blocks, block_size, num_kv_heads, head_size)` and is returned by https://github.com/torch-spyre/spyre-inference/blob/7c4412b40e4047a859b9bc402c7954450f7f0438/spyre_inference/v1/attention/backends/spyre_attn.py#L137-L146. Can we check that this method here would be compatible to that?

Also, in the model runner, we will likely allocate the KV cache tensor in the ModelRunner, i.e., here: https://github.com/torch-spyre/spyre-inference/blob/7c4412b40e4047a859b9bc402c7954450f7f0438/spyre_inference/v1/worker/spyre_model_runner.py#L303-L308. Would it make sense to put the tensor allocation there already?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,591 @@
...
+        assert attn_metadata.num_seqs == 1, (
+            f"Spyre attention requires num_seqs=1, got {attn_metadata.num_seqs}"
+        )
+
+        num_actual_tokens = attn_metadata.num_actual_tokens
+
+        if self._k_cache_dev is None:
+            self._init_device_cache(kv_cache.shape[0], attn_metadata.block_size)
+
+        # Step 1: Scatter — mask is precomputed by builder, only fill values per layer
```
bohnstingl: Can we still make a comment which part and which tensor is expected to be on spyre and which on CPU? Maybe even indicate the device in the name for the moment? For example, `_spyre` and `_cpu`? All the buffers are still on CPU I suppose, right?

spyre_inference/platform.py
```diff
@@ -46,10 +47,15 @@ class TorchSpyrePlatform(CpuPlatform):
     dispatch_key: str = "CPU"
 
     # Register the PyTorch Native Attention implementation as the CUSTOM backend
-    register_backend(
-        AttentionBackendEnum.CUSTOM,
-        "spyre_inference.v1.attention.backends.spyre_attn.SpyreAttentionBackend",
-    )
+    _SPYRE_ATTN_IMPL = os.environ.get("SPYRE_ATTN_IMPL", "default")
+
+    if _SPYRE_ATTN_IMPL == "exp":
```
joerunde: Do we intend to keep these two backends toggle-able, or just use this new one as the default?

If we want to keep both around then I think we'd need to parameterize the local attention tests (and inject parameterization for the upstream attention tests?) to run both with and without this set.

But unless there's a good reason to keep both, I'd prefer to just pick one and keep the configuration surface to a minimum here

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,591 @@
...
+    KV_LENGTH_ALIGNMENT (256), bucketing sequence lengths to reduce
+    torch.compile recompilations.
+"""
+
+from dataclasses import dataclass
+from typing import ClassVar
+
+import torch
+
+from vllm_spyre_next.custom_ops.utils import convert
```
joerunde: this needs fixing

joerunde (changes requested):
After rebasing on main and fixing the import errors locally I can manually enable this and see passing `test_spyre_attn` tests. However the upstream tests fail because they use batch sizes > 1.

If we want to merge this, we'd need to both override the upstream tests to only run batch size 1 for this attention backend, and put a check in `platform.py` that disallows `--max-num-seqs > 1` for this attention backend

spyre_inference/platform.py
tdoublep: I also tend to think it is better to just replace it with this one (even though it might be slower). This attention backend is functionally much closer to what we want, since the KV cache is residing on Spyre. We can then iteratively improve it from here.

jvlunteren: I agree. I will replace the original backend with the new one.

spyre_inference/v1/attention/backends/spyre_attn_exp.py
jvlunteren: Fixed.

tdoublep:
Now that this replaces the backend, should we just have platform.py override max-num-seqs to 1 for the time being?

joerunde:
@tdoublep yeah that'd be required. I think we'll also need to update an upstream test fixture to not run the attention tests with more than batch size 1

joerunde:
bot:next-test

joerunde:
all of the attention tests are failing, see internal CI logs

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -212,360 +405,167 @@ def __init__(
...
             query_per_seq, compact_k, compact_v, mask, query.device, query.dtype
         )
 
-        # Step 6: Extract only the actual query tokens (strip padding) (CPU)
-        # [num_actual_tokens, num_heads, head_size]
-        attn_output_flat = self._extract_relevant_output(attn_output, attn_metadata.query_start_loc)
-
-        output[:num_actual_tokens].copy_(attn_output_flat)
+        # Step 6: Extract actual tokens (num_seqs=1 → simple slice)
+        output[:num_actual_tokens].copy_(attn_output[0, :num_actual_tokens])
```
bohnstingl: Can we try to have the output of the attention to reside on spyre? This would better fit the flow of the SpyreModelRunner and would simplify things in the following MLP layers.

bohnstingl (changes requested):
Can we try to have the final attention output on spyre?

Also, are there any thoughts on my previous comments? Especially the allocation of the KV cache tensors. Shall we try to allocate them in the proper `_allocate_kv_cache_tensors` in the `SpyreModelRunner`, as upstream vLLM also does? I am happy to help on this if we decide to do it.

spyre_inference/v1/attention/backends/spyre_attn_exp.py
bringlein: > Can we check that this method here would be compatible to that?

@bohnstingl why do you think that is important? If we don't have other attention layers in the beginning? I think we may have to change the format in the end to be not `num_blocks, 2, block_size, num_kv_heads, head_size) or (2, num_blocks, block_size, num_kv_heads, head_size` (still up to discussion, see other issues). Would be good to know any dependency on the layout (besides other attention layers) early.

bohnstingl: Okay, we don't necessarily have to force any specific shape at the moment, I believe. The only thing that we should do is to allocate the (multiple) KV cache tensors in the right place, i.e., the `SpyreModelRunner` via the `_allocate_kv_cache_tensors` function. I briefly discussed this with @tdoublep and we will do this in a follow-up PR. WDYT?

bringlein: sounds good to me!

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test
(hoping test cluster is back operational now)

joerunde:
bot:next-test

jvlunteren:
bot:next-test

joerunde:
bot:next-test
(@jvlunteren I just added you to the internal allow-list so the comments should work for you now)

jvlunteren:
bot:next-test

jvlunteren:
bot:next-test

jvlunteren:
bot:next-test

spyre_inference/platform.py
bohnstingl: Can we reflect the removal of the older backend with the new one in this PR?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,741 @@
...
+        """
+        # Compact per-layer transfer: [num_tokens, num_kv_heads, head_size] on Spyre
+        k_new_dev = (
+            key.to(self._target_dtype).to(self._target_device).contiguous()
+        )  # [num_tokens, num_kv_heads, head_size]
+        v_new_dev = value.to(self._target_dtype).to(self._target_device).contiguous()
+
+        import os as _os_ow
+
+        if _os_ow.environ.get("SPYRE_SCATTER_USE_OVERWRITE", "0") != "0":
```
bohnstingl: I don't think the import alias is necessary here? I think we can just do 

```
import os

_SPYRE_SCATTER_USE_OVERWRITE = os.environ.get(
    "VLLM_SPYRE_SCATTER_USE_OVERWRITE", "0"
) != "0"
```

In the longer run, if we device to keep this section or any other environment-variable based decision, maybe we should think about adding a custom env.py and override the upstream one?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,741 @@
...
+            # k_new_dev[t, :, :] (shape [num_kv_heads, head_size]) into
+            # cache[:, block_indices[t], block_offsets[t]*head_size :
+            #       block_offsets[t]*head_size + head_size]
+            # via overwrite_f with dims=[1, 2], offsets=[br, cs].
+            # Input must be [num_kv_heads, 1, head_size] to match the
+            # output's rank (3D) with the listed dims being singleton.
+            head_size = self.head_size
+            block_size = self._block_size
+            assert block_size is not None
+            sm = slot_mapping.detach().to("cpu")
```
bohnstingl, line 634: I think the `.detach()` is not necessary?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,741 @@
...
+            block_indices = (sm // block_size).tolist()
+            block_offsets = (sm % block_size).tolist()
+            num_tokens = k_new_dev.shape[0]
+
+            # Build per-token inputs on CPU and upload as fresh tensors.
+            # Slicing k_new_dev[t] produces a Spyre view whose device
+            # layout doesn't match what overwrite_f's bundle compiler
+            # expects (triggers patch (A)'s narrowed-view guard). A
+            # fresh CPU->Spyre upload reshapes cleanly.
+            k_cpu = key.to(self._target_dtype).contiguous()
```
bohnstingl, line 644: I feel like there are a lot of `.contiguous()` calls. Are all of them really necessary? I thought that they are mainly used to trigger restickification after a split or indexing operation?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -0,0 +1,741 @@
...
+                k_tok = k_cpu[t].unsqueeze(1).contiguous().to(self._target_device)
+                v_tok = v_cpu[t].unsqueeze(1).contiguous().to(self._target_device)
+                k_cache = torch.ops.spyre.overwrite_f(
+                    input=k_tok, output=k_cache, dims=[1, 2], offsets=[br, cs]
+                )
+                v_cache = torch.ops.spyre.overwrite_f(
+                    input=v_tok, output=v_cache, dims=[1, 2], offsets=[br, cs]
+                )
+            self._k_cache_dev = k_cache
+            self._v_cache_dev = v_cache
```
bohnstingl, line 661: If the `torch.ops.spyre.overwrite_f` allows to provide an output tensor, there are two questions arising in my mind:
1. Do we need to collect the results, or wouldn't the output argument simply be enough? For example, instead of `k_cache = torch.ops.spyre.overwrite_f(... output=k_cache, ...)`, just `torch.ops.spyre.overwrite_f(... output=k_cache, ...)`?
2. Can't we use the `self._k_cache_dev` directly instead of creating the `k_cache` tensor?

bohnstingl (changes requested):
Thank you @jvlunteren for the updated PR! I went through and left some comments.

joerunde:
> Can we reflect the removal of the older backend with the new one in this PR?

@bohnstingl I don't think the new experimental one is ready yet, hence why we went back to leaving this gated behind an environment variable until we get it passing all tests.

I still see the upstream attention tests failing for example:
```
SPYRE_ATTN_IMPL=exp pytest tests -m "upstream and attention" -vx
...
E           AssertionError: [AttentionBackendEnum.CUSTOM] output differs from SDPA baseline. Tensor-likes are not close!
E           
E           Mismatched elements: 3436 / 4096 (83.9%)
E           Greatest absolute difference: 0.25341796875 at index (0, 19, 12) (up to 0.01 allowed)
E           Greatest relative difference: 1.005859375 at index (0, 20, 90) (up to 0.01 allowed)
```

(also we'd need the changes back that force batch size 1 too)

joerunde:
bot:next-test

joerunde:
@jvlunteren @bohnstingl I pushed a couple small commits to:
1. Re-enable the attention tests that I previously marked with skips because the experimental backend failed on them
2. Extract the `os.getenvs` to an `envs.py` because it's a very simple and easy-to-claude change to consolidate our getenv usage

I think we can go ahead and merge this- the idea is to close out this old PR and get the code to main so that we can start iterating on the on-device attention backend together. Once it's ready to be enabled by default, we can delete the existing backend and feature gate

joerunde (approved):
I'm okay merging this to get the experimental backend to main for further iteration

jvlunteren:
bot:next-test

bohnstingl (approved):
I went through the code again. It is functional and to me it is fine to merge this as-is and iterate on it. I will prepare a draft PR with my suggestions.

tdoublep (approved):

bringlein (approved):
LGTM

Issue #121, opened 2026-01-23
Paged KV-cache implementation of AttentionBackend using torch-spyre


tdoublep:
Follow-on from torch-spyre/spyre-inference#122 

- Implement a PyTorch-native paged attention backend for vLLM that is compatible with torch-spyre.
- The FMS attention backend from torch-spyre/spyre-inference#123 may serve as a starting point
- Extend the native code with any annotations that may be required into order to achieve good performance.
- These annotations may eventually take the form of some tiled DSL (e.g., Helion)

bringlein:
Based on the code @bohnstingl created [here](https://github.com/bohnstingl/vllm/tree/naive_attn_backend) (and see comment in torch-spyre/spyre-inference#122), we are close to get a "pytorch-native paged attention backend". 
Next open questions:
1. Currently it depends on the order of blocks, we have to verify if this assumption holds in all cases. If not, can we enforce this in vllm or do we need to change the kernel?
2. integrate in the `vllm_spyre_next` plugin
3. ops coverage: What would be better/worse alternatives?

bohnstingl:
I opened a draft PR which extends the vllm_spyre_next plugin with the option to integrate new custom attention backends and also attached a usage example to it, see https://github.com/vllm-project/vllm-spyre/pull/774

bringlein:
Related to torch-spyre/spyre-inference#80 but also for the broader picture here: Would it be an option to write a pytorch native backend that has a similar workaround like the one that is currently used: To convert the indirect memory accesses to accesses with offsets? 
I.e. converting the page table to a tensor with static offsets that can be compiled? 
This would be similar to recording a cuda graph for a fixed batch size where the kernel could accesses more pages than in the request and hen would need to know how much tokens are valid.

tdoublep:
PR with this backend: https://github.com/vllm-project/vllm-spyre/pull/774

bohnstingl:
I have been updating the PR, targeting `vllm-spyre-next`, to provide an initial skeleton for the various attention implementations we might have for torch-spyre. Issue https://github.com/vllm-project/vllm-spyre/pull/774 has been closed in favor of https://github.com/vllm-project/vllm-spyre/pull/798, which is the current issue now to track the progress.

tdoublep:
I am moving this to Backlog since we need to focus 100% on the contiguous attention implementation for 1Q.

tdoublep:
[...]

Issue #80, opened 2026-02-18
Find a way to used indirect indexing during the attention computation


bohnstingl:
The naive implementation [here](https://github.com/bohnstingl/vllm/blob/bb0782267790fc9c479286e4772b28b7925bf34d/vllm/v1/attention/backends/pytorch_native_attn.py#L427-L473) uses the entire KV cache to compute the attention. This is not performant and to implement the proper paged attention, a way to only compute with the relevant blocks is required.

tdoublep:
We now have support for list-based paged attention on main since #205 was merged.

Issue #20, opened 2026-04-15
verify indirect memory access with absolute addresses


bringlein:
### To discuss

as proposed in torch-spyre/spyre-inference#121 

### Checklist

- [x] I have searched for similar feature requests or RFCs.

bringlein:
I experimented a bit and as discussed (see https://github.com/torch-spyre/spyre-inference/issues/121#issuecomment-3921070695) , having the paged memory organized as lists, works: https://github.com/torch-spyre/spyre-inference/pull/5

bringlein:
(and thanks @sducouedic for helping me to set up the new container)

tdoublep:
@tdoublep to review PR

bringlein:
discussed with torch-spyre team and created yet another PR with a script that is not yet working, but states the _desired_ behavior: #147

tdoublep:
@bringlein tries to verify this but it doesn't work. We can't use absolute addresses right now since we can't get them. May revisit once we can get access to these addresses.

PR #133, opened 2026-04-24
testing indirect memory access via .data_ptr()

bringlein:
See also discussion in https://github.com/torch-spyre/spyre-inference/issues/121#issuecomment-4307914959  (also related: https://github.com/torch-spyre/spyre-inference/issues/80 and #20 ).


bringlein:
Current limitation with this approach: `.data_ptr()` returns an integer larger than int32, and we can only transfer int32 tensors to the device, currently.

PR #140, opened 2026-04-27
:memo: Add .gitignore and README

joerunde:
This seeds a .gitignore file, and adds an initial README because our existing one is busted and we'll want one to start pushing to pypi with.

Also adds a `CLAUDE.md` file that's been working well for me recently


.gitignore
```diff
@@ -0,0 +1,45 @@
...
+*.cover
+
+# probable secrets
+.env
+
+# uv
+uv.lock
+
+# CLAUDE.md (AI-generated, not source)
+CLAUDE.md
```
tjohnson31415: I think it is more typical to include CLAUDE.md (or AGENTS.md more generically) in .git to provide shared repo-sepecific guidance for the AI.

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+.pytest_cache/
+.coverage
+htmlcov/
+*.cover
+
+# probable secrets
+.env
+
+# uv
+uv.lock
```
tjohnson31415: Uh, isn't the whole point of uv.lock to commit it?

README.md
tjohnson31415: Is this `sendnn` comment correct? Or should we just say something like "Access to IBM Spyre hardware with the Spyre Runtime stack" (e.g. I think more than just `sendnn` is required)

tjohnson31415:
```suggestion
The plugin automatically registers with vLLM when installed.
```

tjohnson31415: I don't know why the diff view doesn't show it, but this suggested change is to remove "Use it by setting the appropriate platform configuration:" since there is no "platform configuration" to set. Though we could recommend setting VLLM_PLUGINS here.

.gitignore
joerunde: not according to bootleg claude lol

🤦

.gitignore
joerunde: True, if we're okay with that then I'll commit what I have

tjohnson31415 (approved):
LGTM

PR #141, opened 2026-04-27
:test_tube: Setup custom test fixtures for upstream attention tests

joerunde:
Pulling changes over from https://github.com/torch-spyre/sendnn-inference/pull/884

This isn't quite working as is, looks like maybe the shape of the attention output changed since this was last worked on.

Update: This is working now after also applying a patch in the fixture to transpose the attention result back to the layout that the vllm tests expect.


joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
This seems to be randomly failing with segfaults, which is not ideal but I'm not sure if it's our fault or torch-spyre's

joerunde:
bot:next-test

tjohnson31415:
bot:next-test

tjohnson31415:
bot:next-test

tjohnson31415:
Using these changes, I was able to get a full passing test run in a dev pod (`VLLM_COMMIT=v0.19.1 pytest -svx tests/`) with the `dev-next-test` image.

The segfaults on the base test image may be a known issue that has been fixed in the new `dev-next-test` base image, but the new `dev-next-test` image requires an update to torch-spyre...
And there's a confounding factor that UV caching of torch-spyre compilation is not aware of system dependencies (only the git commit hash is used to key the cache). The first CI run on `dev-next-test` failed tests with an internal assertion error and the second run (after I cleard the cache) had the expected compilation failure.

So maybe we just get this PR reviewed and merged and worry about full passing CI tests in a separate PR updating torch-spyre / base image.

spyre_inference/testing/pytest_plugin.py
```diff
@@ -598,6 +627,72 @@ def should_do_global_cleanup_after_test():
...
+
+    # _test_backend_correctness may be called with a hardcoded AttentionBackendEnum.FLASH_ATTN,
+    # which we want to ignore
+    orig_tbc = test_module._test_backend_correctness
+
+    def tbc_wrapper(
+        batch_spec, model, backend_to_test: list[AttentionBackendEnum | str], *args, **kwargs
+    ):
+        if "AttentionBackendEnum.FLEX_ATTENTION" in str(backend_to_test):
+            return
```
tjohnson31415: The comment says we want to ignore `FLASH_ATTN`, but the code ignores `FLEX_ATTENTION`

spyre_inference/testing/upstream_tests.yaml
```diff
@@ -33,17 +36,29 @@ tests:
     #           hidden_size: [64]
     #   block_list:
     #     - test: "test_fused_rms_norm_quant"
-    - rel_path: tests/models/language/generation/test_common.py
+
+    # Model tests are now failing too :(
```
tjohnson31415, line 40: What's the context on this comment?
Did these only start failing with this PR, or it just wasn't noticed before?

joerunde: TBH I don't really know, I need to get back up to speed on all the open issues here. But I don't think injecting fixtures for other tests broke these

joerunde:
@tjohnson31415 I'm with you there- want to go ahead and approve for merging?

Also very 😭 about the uv caching- that shaves off multiple minutes from the test run...

tjohnson31415 (approved):
LGTM

PR #143, opened 2026-04-28
Add Dockerfile for building vllm-spyre on x86 

tapanbhoi:
## Summary
Add Dockerfile for building vllm-spyre on x86 (amd64).

## Changes
- Added dockerfile/Dockerfile.amd64

## Validation
- Build tested on x86 machine
- Image builds successfully
- Container runs and environment is ready

### DIFF
--- a/dockerfile/Dockerfile.amd64
+++ b/dockerfile/Dockerfile.amd64
@@ -1,0 +1,18 @@
+FROM 
+
+USER root
+
+ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_VLLM_SPYRE_NEXT=0.1.0
+
+RUN cd /root && \
+    git clone --depth 1 https://github.com/torch-spyre/spyre-inference.git && \
+    source /root/torch-spyre/.venv/bin/activate && \
+    uv pip install vllm torchvision && \
+    cd spyre-inference && \
+    uv pip install -e . --no-deps && \
+    echo "source /root/torch-spyre/.venv/bin/activate" >> /root/.bashrc
+
+USER senuser
+WORKDIR /root/spyre-inference
+
+CMD ["bash"]


dockerfile/Dockerfile.amd64
```diff
@@ -0,0 +1,18 @@
+FROM icr.io/ai_sw_accel/mayurtorch-spyre:latest
```
joerunde: I don't think we can expose internal image names like this

dockerfile/Dockerfile.amd64
```diff
@@ -0,0 +1,18 @@
...
+
+USER root
+
+ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_VLLM_SPYRE_NEXT=0.1.0
```
joerunde, line 5: We have proper versioning via setuptools_scm, we should not override versions

dockerfile/Dockerfile.amd64
```diff
@@ -0,0 +1,18 @@
...
+
+RUN cd /root && \
+    git clone --depth 1 https://github.com/torch-spyre/spyre-inference.git && \
+    source /root/torch-spyre/.venv/bin/activate && \
+    uv pip install vllm torchvision && \
+    cd spyre-inference && \
+    uv pip install -e . --no-deps && \
```
joerunde, line 12: The proper install procedure should just be `uv sync --frozen` with no other setup or configuration required.

There are install docs with more info here: https://github.com/torch-spyre/spyre-inference/blob/main/docs/getting_started/installation.md

dockerfile/Dockerfile.amd64
```diff
@@ -0,0 +1,18 @@
...
+    echo "source /root/torch-spyre/.venv/bin/activate" >> /root/.bashrc
+
+USER senuser
+WORKDIR /root/spyre-inference
+
+CMD ["bash"]
```
joerunde, line 18: What's the intent of this image? If it's to run vllm, then `vllm serve` should be the entrypoint. If it's to set up a container for development and testing, then we would need an init process like `tini` to reap processes and forward signals

joerunde (changes requested):
needs DCO and more info / documentation on what this image is for and why we should be building it

tapanbhoi:
I am closing the PR as need to finalise the location of internal image location and other dependencies .Will raise the PR agsin.

Issue #46, opened 2026-03-09
Verify numerical equivalence of the SiluAndMul op


bohnstingl:
### Feature description

This issue tracks the progress of ensuring that the SiluAndMul wrapped for torch-spyre is numerically equivalent to the upstream vLLM implementation.

### Motivation and context

NA

### Checklist

- [x] I have searched for similar feature requests

romitjain:
@bohnstingl Can I take a stab at this?
My plan is similar to torch-spyre/sendnn-inference#837 by writing tests in vllm-spyre-next and enabling upstream tests.

cc: @tdoublep

bohnstingl:
@romitjain of course, please take it.
Hopefully this issue is not so involved as the rms_norm one, because we can at least observe that the first couple of tokens of the model with SiluAndMul on `cpu` and on `sp`yre are identical.
I assigned you and move to "In progress".

bohnstingl:
We do have a test for the SiluAndMul operation, see https://github.com/torch-spyre/spyre-inference/pull/144

PR #144, opened 2026-04-29
SpyreSiluAndMul Tests

romitjain:
## Description

This PR adds 2 tests for `SpyreSiluAndMul`: Numerical correctness and OOT plugin compatibility. It also adds upstream tests (commented out). However, upstream tests currently fail due to a lack of `torch.randn` support on spyre hardware. `torch.randn` is used in upstream tests to create a tensor on the device.

## Related Issues

https://github.com/torch-spyre/spyre-inference/issues/46

## Test Plan

```
pytest -rA -m "siluandmul"
```

## Checklist

- [x] I have read the [contributing guidelines](https://docs.vllm.ai/projects/spyre/en/latest/contributing)


romitjain:
This is the port of: https://github.com/torch-spyre/sendnn-inference/pull/863

spyre_inference/testing/upstream_tests.yaml
```diff
@@ -33,6 +37,24 @@ tests:
     #           hidden_size: [64]
     #   block_list:
     #     - test: "test_fused_rms_norm_quant"
+    # Currently the tests fail because upstream rtol and atol are 0
```
joerunde: This is a bit confusing, do they pass with the overrides set here?

joerunde:
bot:next-test

joerunde:
@romitjain this is currently failing for me because the `default_vllm_config` fixture override doesn't set a `ModelConfig`, and both our `apply_config_platform_defaults` and `check_and_update_config` require that to exist. 
I have a quick solution to that in#141, which you could rebase this on once that merges so we can get passing tests. (But feel free to make it better!)

For our local tests (/home/senuser/spyre-inference/tests/test_silu_and_mul.py), I would suggest against trying to pull in the fixture from upstream vllm tests anyway to keep those decoupled. Using our own fixture for those tests would allow us to scope it to the module, which will be much faster since setting up a model config with a simple `ModelConfig()` still takes a bit of time to go through vllm's defaulting and validation bits. 

Once I fixup the `default_vllm_config` problems and delete the broken attention test file, this passes for me so we should be good to merge after rebasing past #141

spyre_inference/testing/upstream_tests.yaml
romitjain: I forgot to remove this

romitjain:
@joerunde 
> For our local tests (/home/senuser/spyre-inference/tests/test_silu_and_mul.py), I would suggest against trying to pull in the fixture from upstream vllm tests anyway to keep those decoupled. 

Since the pytest plugin is installed, it would override the upstream `default_vllm_config` fixture with our fixture.

romitjain:
bot:next-test

romitjain:
bot:next-test
MARKERS="siluandmul"

romitjain:
@joerunde Can you please review the PR now?

The pipeline with just the `siluandmul` marker is passing (`#7`), but the full sweep is failing for me with seg fault in attention tests (`#6`).

tests/test_silu_and_mul.py
```diff
@@ -0,0 +1,88 @@
...
+    "dtype",
+    [
+        pytest.param(
+            torch.float16,
+            marks=pytest.mark.xfail(reason="Strided tensors on float16 does not work on spyre"),
+        ),
+        torch.float32,
+    ],
+)
+def test_spyre_siluandmul_matches_reference(default_vllm_config, num_tokens, d, dtype):
```
joerunde, line 35: Actually, do we need to keep this test around?

The only difference I see is that this uses odd values of `d`. Do those still fail upstream, and if so can we patch with a custom fixture to get them working?

I can go ahead and approve this PR for merging and we can follow up with that later if we want

joerunde (approved):

joerunde:
bot:next-test

joerunde:
bot:next-test

joerunde:
Full sweep is passing, I'll just go ahead and merge

tests/test_silu_and_mul.py
romitjain: Yes @joerunde 
This can be removed. I can follow up in a separate PR to remove this

PR #146, opened 2026-04-29
Rework init_device in TorchSpyreWorker to avoid double-creation of ModelRunner

bohnstingl:
In the current implementation, the `ModelRunner` is created twice. Once during the call to `super().init_device()` and then once again during https://github.com/bohnstingl/spyre-inference/blob/9486493b71496b48df6d95e0dea342fe797f5216/spyre_inference/v1/worker/spyre_model_runner.py#L174-L175. This PR removes the call to `super().init_device()` and overtakes all upstream instructions, except the call to https://github.com/vllm-project/vllm/blob/33f36d42605a476a09ed75936e7c931cb8b432c5/vllm/v1/worker/cpu_worker.py#L140-L143

`torch-spyre` currently has a limitation with work division of matmuls. This requires the weight matrix of the `ParallelLMHead` to be of a specific shape, see https://github.com/torch-spyre/torch-spyre/issues/1918.


spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -41,8 +46,46 @@ def __init__(
...
+        # Note: unique identifier for creating allreduce shared memory
+        os.environ["VLLM_DIST_IDENT"] = self.distributed_init_method.split(":")[-1]
+        # Initialize the distributed environment.
+        init_worker_distributed_environment(
+            self.vllm_config,
+            self.rank,
+            self.distributed_init_method,
+            self.local_rank,
+            current_platform.dist_backend,
+        )
```
bohnstingl: These things are specific for the CPUWorker. Since our SpyreRunner derives from CPU, it would make sense for me to keep them, but does anybody else have comments?

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -41,8 +46,46 @@ def __init__(
         register_all()
 
     def init_device(self) -> None:
-        # Call the upstream init_device function for the environment setup
-        super().init_device()
+        # Call the relevant parts from the CPUWorker upstream, but don't create the CPUModelRunner
```
joerunde: Copy-pasting upstream code is pretty brittle, if the goal is to avoid the double-creation of a model runner then why not just patch out the cpu model runner to not build itself?

```
import vllm.v1.worker.cpu_worker as cpu_worker_module
...
                                
      original = cpu_worker_module.CPUModelRunner                                                
      cpu_worker_module.CPUModelRunner = lambda *a, **kw: None                                   
      try:                                                                                       
          super().init_device()
      finally:                                                                                   
          cpu_worker_module.CPUModelRunner = original
                                                                                                 
      self.model_runner = TorchSpyreModelRunner(                                                 
          self.vllm_config,
          torch.device("spyre"),                                                                 
      )
```

spyre_inference/v1/worker/spyre_worker.py
joerunde: We don't support any distribution today so I don't think it matters all that much, and IIUC our plan is to eventually stop deriving from the cpu model runner / worker once we fully support spyre, right?

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: @joerunde good catch. I changed the code for a patching-based approach, instead of copying the upstream code

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -1,10 +1,15 @@
 """A Torch Spyre worker class."""
 
+import os
+import sys
```
yannicks1: unused imports

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -1,10 +1,15 @@
...
 import torch
 
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
 from vllm.v1.worker.cpu_worker import CPUWorker
+from vllm.v1.worker.gpu_worker import init_worker_distributed_environment
+from vllm.utils.torch_utils import set_random_seed
+from vllm.platforms import CpuArchEnum, current_platform
```
yannicks1: unused imports

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -41,14 +46,21 @@ def __init__(
         register_all()
 
     def init_device(self) -> None:
-        # Call the upstream init_device function for the environment setup
-        super().init_device()
-
-        # Construct Spyre model runner with torch.device("spyre")
-        self.model_runner = TorchSpyreModelRunner(
+        import vllm.v1.worker.cpu_worker as cpu_worker_module
```
yannicks1: I am always unsure when imports should be at the top of the file vs within functions. @joerunde @maxdebayser @tjohnson31415 what are best practices?

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -41,14 +46,21 @@ def __init__(
...
+                                
+        # Patch the CPUModelRunner with the TorchSpyreModelRunner
+        original = cpu_worker_module.CPUModelRunner
+        cpu_worker_module.CPUModelRunner = lambda *a, **kw: TorchSpyreModelRunner(
             self.vllm_config,
             torch.device("spyre"),
         )
+        try:
+            # We will invoke the upstream init_device method with the
+            # CPUModelRunner patched. This will ensure that everthing for the CPUWorker is setup,
```
yannicks1: everthing -> everything

yannicks1 (commented):
looking good, please address the minor comments and revise the PR description with the patching approach opted for.

spyre_inference/v1/worker/spyre_worker.py
joerunde: My take is they should always be at the top of the file unless they're required to be imported lazily to avoid an import loop. In which case, we should at least document it with a comment but preferably fix the file structure to avoid the loop

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: This has been reworked by just calling the `init_device` function from upstream.

> and IIUC our plan is to eventually stop deriving from the cpu model runner / worker once we fully support spyre, right?

This is not entirely casted in stone yet, but potentially.

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: Removed.

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: Removed.

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: In this case it shouldn't really matter, but I moved it to the top of the file.

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: Done.

joerunde:
I don't think the removed imports were actually removed? (Unless github is lying to me and serving stale commits)

shouldn't be an issue once #169 is merged

bohnstingl:
bot:next-test

bohnstingl:
@joerunde somehow my commit was not accepted. I recommitted and merged the main in as well

bohnstingl:
bot:next-test

tdoublep (approved):
LGTM

bohnstingl:
bot:next-test

joerunde (approved):

PR #147, opened 2026-04-29
Paged Attention example with DESIRED functionallity

bringlein:
The script in this PR shows the DESIRED way of accessing paged memory in spyre, so that it is compatible with vLLM and no further workarounds would be required. The script does not work right now, but hopefully contributes to the discussion, see https://github.com/torch-spyre/torch-spyre/issues/866 and https://github.com/torch-spyre/spyre-inference/issues/121


kiszk:
@bringlein According to my understanding, this example tries to do followings:
- Calculate actual memory offsets from `page_table_list` and store them into `abs_page_table_K` and `abs_page_table_V`
- Access a subset of `a_pages` and `b_pages` using `abs_page_table_K` and `abs_page_table_V`
IIUC, it looks like writing an assembler code.

But, I think there are some mistakes in this implementation.
A compiler translates a given index in `[]` into actual offsets by using a type and/or shape of variable (e.g. `a_pages`). In the case of `a_pages`, its type and shape are `torch.Tensor` and `torch.Size([512, 16, 8, 128])`. I think that a value of `a_page_table[0]` is very huge (at least  more than 512 since this element is expected to keep an actual memory offset calculated by stride). So, an access `a_pages[a_page_table[i]]` is incorrect.

I think that code can be written simply by using `torch.index_select` that can include `for i in range(max_page_table_length):`. Of course, `torch.index_select` should be supported by torch-spyre. I will try to rewrite an example after national holidays.

@tardieu any comments?

bringlein:
Thanks for you questions @kiszk. 

> ...it looks like writing an assembler code.

..well, we would also wish to use less workarounds...(for the time being)

>  think that a value of a_page_table[0] is very huge

I'm not sure I following. Yes, we want to have a full page (of e.g. 16 tokens) to be accessed at once, so for me, the indexing is as we want it to. 

> I think that code can be written simply by using torch.index_select that can include for i in range(max_page_table_length):. Of course, torch.index_select should be supported by torch-spyre. 

I think here is maybe another misunderstanding: One very important detail is to **avoid** creating new tensors in DRAM/HBM, see discussion in https://github.com/torch-spyre/torch-spyre/issues/866#issuecomment-4313976864. According to the [official documentation of index_select](https://docs.pytorch.org/docs/2.11/generated/torch.index_select.html), it creates a **new** tensor. So, if I'm not mistaken, it would be precisely the behavior we have to avoid. 

I think it was discussed before, but it is _critical_ for the performance of paged attention to just work with views and true indirect access. Copying paged memory to new continuous memory is a) what we can already do and b) has horrible performance. 

> I will try to rewrite an example after national holidays.

looking forward to it, I think it will help the discussion. But first, enjoy your holidays @kiszk !

bringlein:
One more comment in this context to have the full picture: If there _would_ be a version of `torch.index_select` that creates a *view* instead of a *new* tensor, it would also enable true paged attention. 
However, this is not supported in pytorch upstream ([and never will be](https://discuss.pytorch.org/t/index-select-same-storage-as-the-original-tensor/4368)) and therefore it would then be a torch-spyre-specific op. 

As context: Having an `index_select_view` would allow kernel authors to have similar functionality like `tl.load`/`hl.load` in Triton/Helion kernels: It would create objects in shared/local memory that are independent of the strides in DRAM/HBM. Hence, if we _could_ create tensors only in the lx scratchpad of Spyre, `torch.index_select` would be useful.

However, to be clear: Having true indirect access via `view_tensor = tensor[absolute_address_tensor[i]]` (or a similar API) is IMHO the fastest way to unblock paged attention (also, if we consider the limited space of lx, having views instead of copies is preferable, IMHO).

kiszk:
@bringlein Thanks for your explanation. I got your point regarding `torch.index_select`. From the performance perspective, I agree with you that it does not create a new tensor.

experimental/attention_backend/paged_vector_add_target.py
```diff
@@ -0,0 +1,228 @@
...
+#   attention computation ops enablement....)
+
+print("prepare computation...")
+
+
+def paged_vector_add(
+    a_pages, b_pages, a_page_table, b_page_table, max_page_table_length, out_pages
+):
+    # output pages are starting from 0, in this case
+    for i in range(max_page_table_length):
```
bohnstingl, line 150: Wouldn't the sequential nature of the for-loop be a source for severe performance degradation? Irrespective of whether we pre-compile the function `paged_vector_add` or not, the individual iterations of the for-loop would still be executed sequentially, which is not necessary as they do not have any sequential dependency? I suppose in a triton or CUDA kernel for example, one would launch individual programs for the iterations? Or at least do it in a blockwise parallel?

bringlein: well, that is a complex trade-off and as shown in the blog posts [here](https://pytorch.org/blog/enabling-vllm-v1-on-amd-gpus-with-triton/) or [here](https://pytorch.org/blog/portable-paged-attention-in-helion/), we have multiple dimensions of parallelism (parallel heads, parallel batch, q blocks) but _still_ an inner loop (over the context).  

I want to have the discussion here on the "enablement" side, and hence a simple loop I guess is enough as a test. 

>  I suppose in a triton or CUDA kernel for example, one would launch individual programs for the iterations? Or at least do it in a blockwise parallel?

Yes, but that is a different programming model than for Spyre.

tdoublep (approved):
LGTM

PR #148, opened 2026-04-29
Ensure SiluAndMul layer is compatible with SpyreRunne

GOavi101:
- Slice input on CPU before transferring to Spyre

- Avoids 'free(): invalid size' crash from slicing Spyre tensors directly

- Fixes garbage output in model inference (SwiGLU activation)

- spyre_copy_from handles contiguity during H2D transfer automatically


spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +45,69 @@ class SpyreSiluAndMul(SiluAndMul):
...
-        self._layer_name = register_layer(self, "spyre_siluandmul")
-
-        logger.debug_once(
-            "SpyreSiluAndMul: Dispatch: enabled=%s, Forward method=%s, Compiled=%s",
-            self.enabled(),
-            self._forward_method.__name__,
-            self.maybe_compiled_forward_spyre is not self.forward_spyre,
+        logger.warning_once(
+            "SpyreSiluAndMul: no dtype promotion (torch-spyre limitation), "
+            "expect numerical differences to upstream vLLM.",
```
bohnstingl: I think this comment is not quite accurate and might be a copy-paste error. Upstream is not doing any dtype upcasting, see https://github.com/vllm-project/vllm/blob/b55b26520c088eeed791cbe4648c73ea015f3613/vllm/model_executor/layers/activation.py#L116-L151.

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +45,69 @@ class SpyreSiluAndMul(SiluAndMul):
...
-
-        # Transfer back to original device and restore original dtype
-        return convert(out, x_device, x_dtype)
-
+        # Convert dtype on CPU, slice there, then transfer each half to Spyre.
+        # spyre_copy_from makes non-contiguous slices contiguous during H2D.
+        x_cpu = x.to(device="cpu", dtype=self._target_dtype)
+        d = x_cpu.shape[-1] // 2
+        x1 = x_cpu[..., :d].to(self._target_device)
+        x2 = x_cpu[..., d:].to(self._target_device)
```
bohnstingl: Can we use the `convert` function here?

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +45,69 @@ class SpyreSiluAndMul(SiluAndMul):
...
-            convert(x1, self._target_device, self._target_dtype),
-            convert(x2, self._target_device, self._target_dtype),
-        )
-
-        # Transfer back to original device and restore original dtype
-        return convert(out, x_device, x_dtype)
-
+        # Convert dtype on CPU, slice there, then transfer each half to Spyre.
+        # spyre_copy_from makes non-contiguous slices contiguous during H2D.
+        x_cpu = x.to(device="cpu", dtype=self._target_dtype)
```
bohnstingl: Can we use the convert function here?

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +45,69 @@ class SpyreSiluAndMul(SiluAndMul):
...
-    x: torch.Tensor,
-    output: torch.Tensor,
-    layer_name: str,
-) -> None:
-    """Custom op implementation — runs outside torch.compile graph."""
-    layer = get_layer(layer_name)
-    result = layer._forward_spyre_impl(x)
-    output.copy_(result)
+        out = F.silu(x1) * x2
+        return convert(out, device=x_device, dtype=x_dtype)
```
bohnstingl: This convert should not be necessary. Can you please confirm that the input to this layer, i.e., `x`, is on spyre? I think the only reason why we currently still need converts is because of the slicing operation, right?

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +45,69 @@ class SpyreSiluAndMul(SiluAndMul):
...
-    """Register the spyre_siluandmul custom op with vLLM."""
-    direct_register_custom_op(
-        op_name="spyre_siluandmul",
-        op_func=_op_func,
-        mutates_args=["output"],
-        fake_impl=_fake_impl,
-    )
-    logger.info("Registered custom op: SpyreSiluAndMul")
+    """No-op: the custom-op barrier has been removed."""
+    logger.debug("SpyreSiluAndMul: no custom ops to register")
```
bohnstingl: Can we remove this function and then also remove the call to `register()` in https://github.com/torch-spyre/spyre-inference/blob/7c4412b40e4047a859b9bc402c7954450f7f0438/spyre_inference/custom_ops/__init__.py#L23?
After doing this, can you please check that the `forward_oot` is still invoked though?

bohnstingl (changes requested):
I provided some comments. In general the PR looks good to me, but we could do a little bit of more cleanup

joerunde:
bot:next-test

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +44,64 @@ class SpyreSiluAndMul(SiluAndMul):
...
+        Sets up the target device (Spyre) and dtype (float16) for computation.
+        The simplified implementation computes directly in forward_oot() without
+        requiring layer registry or custom op registration.
         """
         super().__init__(*args, **kwargs)
 
         logger.debug("Building custom SiluAndMul")
 
         self._target_device = torch.device("spyre")
         self._target_dtype = torch.float16
```
bohnstingl: torch-spyre has support for torch.float32, I believe. Thus, I am wondering whether we actually need this anymore? I think it can be removed and we just perform the operations on the `dtype` of the input.

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +44,64 @@ class SpyreSiluAndMul(SiluAndMul):
...
-        mutates_args=["output"],
-        fake_impl=_fake_impl,
-    )
-    logger.info("Registered custom op: SpyreSiluAndMul")
+        x1 = x[..., :d].contiguous()
+        x2 = x[..., d:].contiguous()
+        
+        # Transfer contiguous slices to Spyre with target dtype.
+        x1 = convert(x1, device=self._target_device, dtype=self._target_dtype)
+        x2 = convert(x2, device=self._target_device, dtype=self._target_dtype)
```
bohnstingl: I think here we can just use the device from the input directly. I.e., by doing something similar to the old version

```
x_device = x.device
...
x1 = convert(x1, device=x_device)
```

bohnstingl (changes requested):
In principle looks good to me. Just maybe apply the small code simplifications that I proposed and update the comments. Also, please run the test on your side locally and trigger the bot for testing.

spyre_inference/custom_ops/silu_and_mul.py
```diff
@@ -48,138 +39,45 @@ class SpyreSiluAndMul(SiluAndMul):
...
-        mutates_args=["output"],
-        fake_impl=_fake_impl,
-    )
-    logger.info("Registered custom op: SpyreSiluAndMul")
+        x1 = x[..., :d].contiguous()
+        x2 = x[..., d:].contiguous()
+        
+        # Transfer contiguous slices to original device with original dtype.
+        x1 = convert(x1, device=original_device, dtype=original_dtype)
+        x2 = convert(x2, device=original_device, dtype=original_dtype)
```
bohnstingl: I think the explicit specification of the `dtype` is not necessary, as the `dtype` should not change in any of the above operation.

bohnstingl (approved):
LGMT. I just left a last comment.
Can you confirm that the output of the example model is as expected?

PR #151, opened 2026-04-30
[Spyre-next] Remove custom op barrier for linear layers

### LINKED ISSUE #31: [Feature]: Ensure MLP layer is compatible with SpyreRunner
### Feature description

The aim of this issue is to rework the RowParallelLinear and ColumnParallelLinear layer such that they can consume inputs on `spyre`, create / keep all layer parameters on `spyre` and produce outputs on `spyre`. Ideally all D2H and H2D data transfers should alleviated. 
Furthermore, it is essential to investigate whether the current opaque layer, for example `torch.ops.vllm.spyre_merged_col_linear` and `torch.ops.vllm.spyre_row_parallel_linear`, presenting a barrier for torch.compile can be removed. This would make the entire layer is torch-compileable and offer significant benefits for optimization and efficiency. However, for this to work, the layer implementation on `spyre` needs to work in eager and torch-compile mode, which may not always be true and depends on torch-spyre support.

### Motivation and context

This work is necessary in order to enable a model to run truly E2E on `spyre`, without immediate D2H and H2D transfers.

### Proposed solution

_No response_

### Checklist

R3hankhan123:
Remove the custom op barrier from all linear layer implementations (MergedColumnParallelLinear, QKVParallelLinear, RowParallelLinear) to enable proper torch.compile via spyre model runner

## Tests Ran
Ran inference tests with both eager mode and compile mode.

## Test Results:
Both outputs are the same for eager mode and compile mode and no garbage output is being generated

Resolves Issue #31

### DIFF
--- a/spyre_inference/custom_ops/linear.py
+++ b/spyre_inference/custom_ops/linear.py
@@ -12,8 +12,10 @@ layer classes used inside MLP blocks:
     - SpyreRowParallelLinear          — replaces RowParallelLinear
       (vllm/model_executor/layers/linear.py)
 
-Since tensor_parallel=1 is assumed, both classes are functionally equivalent
-to F.linear(input, weight, bias) and share the same implementation pattern.
+At TP=1, the upstream forward() methods reduce to quant_method.apply() + bias
+handling.  We inject a custom quant_method (SpyreUnquantizedLinearMethod) that
+performs F.linear directly, QKV and RowParallel still override forward()
+for device placement (D2H after GEMM, H2D before GEMM).
 
 Spyre Device Constraints:
     - Computations performed in torch.float16:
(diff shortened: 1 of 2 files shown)


spyre_inference/custom_ops/linear.py
```diff
@@ -58,40 +56,32 @@ def _init_spyre_linear(self, layer_prefix: str):
...
-        out = self.maybe_compiled_forward_spyre(
-            convert(x, self._target_device, self._target_dtype),
-            convert(self.weight.data, self._target_device, self._target_dtype),
-            convert(bias, self._target_device, self._target_dtype) if bias is not None else None,
-        )
+        # Convert input to Spyre device/dtype
+        # Weights and bias are already on Spyre which is moved by model runner
+        x_spyre = convert(x, self._target_device, self._target_dtype)
 
+        out = F.linear(x_spyre, self.weight.data, bias)
```
bohnstingl: Can you please check whether all the converts are necessary here? I think for the `SpyreMergedColumnParallelLinear` and for the `SpyreRowParallelLinear`, the input tensors should already be on spyre and the outputs are also expected to be on spyre. Only the `SpyreQKVParallelLinear`, as it is used in the attention implementation, is currently expected to produce cpu outputs. Can you confirm that, or is the situation different?

If it is the case, then we could remove the `convert` calls here and just move them to `SpyreQKVParallelLinear` directly.

spyre_inference/custom_ops/linear.py
```diff
@@ -101,19 +91,18 @@ class SpyreMergedColumnParallelLinear(SpyreLinearBase, MergedColumnParallelLinea
...
-        else:
-            output = input_.new_empty(
-                input_.shape[0],
-                self.output_size_per_partition,
-            )
-            torch.ops.vllm.spyre_merged_col_linear(input_, output, self._layer_name)
+        """Forward pass for PluggableLayer.
+
+        Args:
+            input_: Input tensor of any device/dtype
```
bohnstingl: I think we should remove the "of any device/dtype".

spyre_inference/custom_ops/linear.py
```diff
@@ -101,19 +91,18 @@ class SpyreMergedColumnParallelLinear(SpyreLinearBase, MergedColumnParallelLinea
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._init_spyre_linear("spyre_merged_col_linear")
+        self._init_spyre_linear()
```
bohnstingl: Can we try to unify these calls? For example, if we change the `_init_spyre_linear(self)` in SpyreLinearBase to `_init_spyre_linear(self, *args, **kwargs)` and here just call `super().__init__(*args, **kwargs)`, it should in theory call both constructors, the one of `SpyreLinearBase` and the one of `MergedColumnParallelLinear`. Could we try that?

spyre_inference/custom_ops/linear.py
bohnstingl: This is a common pattern. Can we consolidate this into `SpyreLinearBase` and here just do `return self._forward_impl(input_)`?

spyre_inference/custom_ops/linear.py
```diff
@@ -127,19 +116,20 @@ class SpyreQKVParallelLinear(SpyreLinearBase, QKVParallelLinear):
...
+
+        Args:
+            input_: Input tensor of any device/dtype
+
+        Returns:
+            Tuple of (output, bias) if return_bias=True, else just output.
+        """
+        output = self._forward_impl(input_)
+        # D2H before downstream .split() — Spyre can't handle strided views
+        output = convert(output, device="cpu")
```
bohnstingl: This convert will likely have to stay, as discussed above

spyre_inference/custom_ops/linear.py
```diff
@@ -153,54 +143,28 @@ class SpyreRowParallelLinear(SpyreLinearBase, RowParallelLinear):
...
+
+        Args:
+            input_: Input tensor of any device/dtype
+
+        Returns:
+            Tuple of (output, bias) if return_bias=True, else just output.
+        """
+        output = self._forward_impl(input_)
+        # Keep output on Spyre — needed for residual add with Spyre hidden_states
+        output = convert(output, device=self._target_device, dtype=self._target_dtype)
```
bohnstingl: Is this convert necessary?

spyre_inference/custom_ops/linear.py
```diff
@@ -153,54 +143,28 @@ class SpyreRowParallelLinear(SpyreLinearBase, RowParallelLinear):
...
-        direct_register_custom_op(
-            op_name=op_name,
-            op_func=_make_spyre_linear_op_func(op_name),
-            mutates_args=["output"],
-            fake_impl=_fake_impl,
-        )
-        logger.info("Registered custom op: %s", op_name)
+    """No-op: the custom-op barrier has been removed."""
+    logger.debug("SpyreLinear layers: no custom ops to register")
+    pass
```
bohnstingl: Can we remove this function and then also remove the call to register() in https://github.com/torch-spyre/spyre-inference/blob/7c4412b40e4047a859b9bc402c7954450f7f0438/spyre_inference/custom_ops/__init__.py#L24

After doing this, can you please check that the forward_oot is still invoked though?

bohnstingl (changes requested):
I provided some suggestions, mostly for code clean-up.

spyre_inference/custom_ops/linear.py
R3hankhan123: Removing the convert function and moving it. to `SpyreQKVParallelLinear` a runtime error is raised
```
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108]   File "/home/rehankhan/vllm-spyre-next/spyre-inference/spyre_inference/custom_ops/linear.py", line 77, in _forward_impl
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108]     return F.linear(x, self.weight.data, bias)
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108]   File "/home/rehankhan/dt-inductor/torch-spyre/torch_spyre/_inductor/decompositions.py", line 286, in __call__
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108]     raise RuntimeError(
(EngineCore pid=18432) ERROR 04-30 12:51:18 [core.py:1108] RuntimeError: Spyre decomposition function called with inputs being on a different device! Args devices: args_device=[device(type='cpu'), device(type='spyre', index=0)]
```

spyre_inference/custom_ops/linear.py
R3hankhan123: Adding some logs in `forward` function I can see its being invoked after removing the register function
```
(EngineCore pid=19187) INFO 04-30 12:57:30 [linear.py:145] SpyreQKVParallelLinear: forward pass executed using spyre model runner
(EngineCore pid=19187) INFO 04-30 12:57:32 [linear.py:170] SpyreRowParallelLinear: forward pass executed using spyre model runner
```

spyre_inference/custom_ops/linear.py
R3hankhan123: Without this i am seeing torch dynamo error

joerunde:
bot:next-test

R3hankhan123:
bot:next-test

R3hankhan123:
bot:next-test

spyre_inference/custom_ops/linear.py
```diff
@@ -12,8 +12,11 @@
     - SpyreRowParallelLinear          — replaces RowParallelLinear
       (vllm/model_executor/layers/linear.py)
 
-Since tensor_parallel=1 is assumed, both classes are functionally equivalent
-to F.linear(input, weight, bias) and share the same implementation pattern.
+At TP=1, the upstream forward() methods reduce to quant_method.apply() + bias
+handling.  We inject a custom quant_method (SpyreUnquantizedLinearMethod) that
+performs F.linear directly, following the same pattern as
+SpyreUnquantizedLMHeadMethod for ParallelLMHead.  QKV and RowParallel still
+override forward() for device placement (D2H after GEMM, H2D before GEMM).
```
bohnstingl: @R3hankhan123 can you please remove the last part of the sentence: ", following the same pattern as SpyreUnquantizedLMHeadMethod for ParallelLMHead". This is not yet merged.

spyre_inference/custom_ops/linear.py
```diff
@@ -22,185 +25,83 @@
     - Tensor parallelism: TP=1 assumed (single Spyre device)
 
 References:
-    - Upstream linear layers: vllm/model_executor/layers/linear.py
-    - Pattern reference:      spyre_inference/custom_ops/rms_norm.py
+    - Upstream linear layers:   vllm/model_executor/layers/linear.py
+    - Pattern reference:        spyre_inference/custom_ops/parallel_lm_head.py
```
bohnstingl: Please remove 
`- Pattern reference:        spyre_inference/custom_ops/parallel_lm_head.py. `
This is not yet merged.

spyre_inference/custom_ops/linear.py
```diff
@@ -22,185 +25,83 @@
...
-        self._layer_name = register_layer(self, layer_prefix)
+        if isinstance(self.quant_method, UnquantizedLinearMethod):
+            self.quant_method = SpyreUnquantizedLinearMethod()
 
         logger.warning_once(
-            "%s: no dtype promotion (torch-spyre limitation),"
+            "%s: no dtype promotion (torch-spyre limitation), "
             "expect numerical differences to upstream vLLM.",
             self.__class__.__name__,
         )
```
bohnstingl: I forgot to remove this comment in my PR. I think upstream linear layers don't do any `dtype` upconversions, right? So this might be an outdated comment to remove.

R3hankhan123:
bot:next-test

spyre_inference/custom_ops/linear.py
```diff
@@ -22,185 +25,83 @@
...
-    ]:
-        direct_register_custom_op(
-            op_name=op_name,
-            op_func=_make_spyre_linear_op_func(op_name),
-            mutates_args=["output"],
-            fake_impl=_fake_impl,
-        )
-        logger.info("Registered custom op: %s", op_name)
+    def forward(self, input_):
+        return super().forward(convert(input_, device=self.weight.device))
```
bohnstingl, line 104: Can you please add a comment here that:
```
RowParallelLinear is currently invoked from `GraniteAttention` where `input_` is on `cpu` and from `GraniteMLP` here `input_` is on spyre. Thus, we always convert the `input_` to `spyre`, which is a NoOp in case of `GraniteMLP`.
```

bohnstingl (approved):
LGTM! Just please add the minor comments. Thanks

joerunde:
bot:next-test

joerunde:
bot:next-test
(just double checking with latest main merged in)

joerunde:
bot:next-test
(test pod failed to start)

PR #153, opened 2026-04-30
⬆️ Changes to bump torch-spyre and vllm

tjohnson31415:
- torch-spyre is not being bumped to the latest, but to a version I have already tested; mainly to get bot:test working quickly
- vLLM 0.20.0 bumps PyTorch 2.11, so holding off on that until torch-spyre/torch-spyre/issues/1315 is resolved
- updating with `uv lock` was getting a bunch of errors in dependency resolution with terratorch, we shouldn't need it for our tests so I updated `sync_upstream_tests.py` to filter it out (tests in dev image still worked without it)


tjohnson31415:
bot:next-test

joerunde (approved):
lpgtm!

PR #157, opened 2026-04-30
♻️ Extract test plugin from spyre-inference package

joerunde:
This PR is a simple refactor to extract the pytest plugin from `spyre-inference/testing` to `tests/plugin`.

This keeps it out of the source tree for `spyre-inference` so we don't accidentally ship it, and decouples the python package management between the two plugins. Upstream test dependency management is now handled in the separate `tests/plugin/pyproject.toml` file, instead of cluttering the root `pyproject.toml`.

This pulls the upstream test dependency management into the plugin as well, and adds it as a standard cli entrypoint to run with `sync-upstream-tests`.


joerunde:
bot:next-test

joerunde:
bot:next-test

tjohnson31415:
bot:next-test

tests/plugin/spyre_testing_plugin/sync_upstream_test_deps.py
```diff
@@ -113,27 +117,32 @@ def filter_requirements(test_in: Path, filtered_libs: set[str]) -> Path:
...
+    skip_line = False
+    for i, line in enumerate(lines):
+        # Detect start of dependencies array
+        if not inside and re.match(r'^dependencies\s*=\s*\[', line):
             inside = True
             depth = line.count("[") - line.count("]")
-            result.append("upstream-tests = []\n")
-            if depth <= 0:
+            # Start fresh dependencies with minimal base
+            result.append('dependencies = [\n    "pytest",\n    "pyyaml",\n')
```
tjohnson31415, line 134: Optional suggestion.
In case we need to add additional plugin (non upstream) dependencies, I think it would be clearer to have a signal comment in the list of dependencies that we auto-populate with upstream deps. Something like:

```
dependencies = [
    "pytest",
    "pyyaml",
    
    # auto-generated upstream dependencies, see sync_upstream_tests.py

```
(could use more obvious sigil in the comment like `# [UPSTREAM VLLM DEPENDENCIES]`)

tjohnson31415 (approved):
One optional suggestion.

Otherwise worked for me in a dev pod and passed in CI!

tjohnson31415:
I also notice that the upstream tests are printed with a leading `::` like
`::test_causal_backend_correctness`.
Another (optional) clarity improvement could be to inject a prefix to the test names, but could be a separate PR. I think we had that at some point via pytest hooks.

joerunde:
bot:next-test

joerunde:
bot:next-test
(there's still one test failure waiting on #161 to fix)

joerunde:
bot:next-test
🤞

PR #159, opened 2026-05-03
[Feature] Wrap ParallelLMHead for sprye

### LINKED ISSUE #7: Wrap the ParallelLMHead for torch-spyre
### Feature description

The goal is to wrap the Wrap the `ParallelLMHead` custom op such that it is compiled for torch-spyre, while the rest of the model remains on cpu. Initially, the layer may not yet be functional on torch-spyre, but the wrapping should be.

### Motivation and context

See https://github.com/torch-spyre/spyre-inference/issues/127

### Proposed solution

_No response_

### Checklist

bohnstingl:
This PR is an attempt to wrap the `ParallelLMHead` layer for use with spyre in such a way that is compatible with the `SpyreModelRunner`. It follows the same pattern as other CustomOp layers, such as RMSNorm, SiluAndMul, etc. However, there are currently two limitations holding this implementation back:

1. There is a bug when executing the linear layer with large shapes that are required for the vocabulary size of a real vLLM model. For example, for the Granite3.3-8B model, the shape of the input is `torch.Size([1, 4096])` and the shape of the weight is `torch.Size([49216, 4096])`. I opened an issue with `torch-spyre` to address this: https://github.com/torch-spyre/torch-spyre/issues/1918.
2. The input to the `ParallelLMHead` still has to reside on CPU. Although the output of the vLLM would be on spyre, they have to be transferred to cpu, because these `hidden_states` get indexed in the GPUModelRunner mutliple times, i.e, https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L4085, https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L4091 or https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L5608


nayana-kumari (approved):
Thanks for this , the implementation looks clean and integrates well with the forward_oot + quant_method flow.

I’m seeing the same issue on my side the CPU → Spyre conversion is manageable, but the large vocab-scale matmul fails with a backend crash, so this seems like a limitation in torch-spyre rather than the wrapping itself.

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,139 @@
...
+        # torch-spyre currently has a limitation with the work division of larger
+        # matmuls. The shapes needs to be a multiple of 64 * (k * 32), where k is
+        # an integer.
+        self.padding = 0
+        pad_1 = self.weight.shape[0] % 64
+        if pad_1 == 0:
+            pad_2 = 32 - (self.weight.shape[0] / 64) % 32
+            if pad_2 > 0:
+                self.padding = int(pad_2 * 64)
+                self.padded_weight = F.pad(self.weight, (0, 0, 0, self.padding))
```
yannicks1: can this happen at init time? `self.weight` is only populated with the actual weights after calling `model_runner.load_model()`. Is it possible that `self.padded_weight` contains uninitialized weights? I do not see any update to `self.padded_weight` after the `load_model()` call.

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,139 @@
...
+        if isinstance(self.quant_method, UnquantizedEmbeddingMethod):
+            self.quant_method = SpyreUnquantizedLMHeadMethod()
+
+        # torch-spyre currently has a limitation with the work division of larger
+        # matmuls. The shapes needs to be a multiple of 64 * (k * 32), where k is
+        # an integer.
+        self.padding = 0
+        pad_1 = self.weight.shape[0] % 64
+        if pad_1 == 0:
+            pad_2 = 32 - (self.weight.shape[0] / 64) % 32
```
yannicks1: I guess using integer division // here instead of float division / is better.

yannicks1: isn't this unnecessarily padding already aligned shapes (which should be zero padded)?

spyre_inference/custom_ops/__init__.py
yannicks1: you removed the `from . import silu_and_mul` in [this](https://github.com/torch-spyre/spyre-inference/commit/4211575c366e40ceff117f8f66ce890384ed2e21#diff-dc213293d5d88ab4109b7e17d3bc8dad6afdcfc7bc8dfd81305e84059bdbcff0) merge commit. Isn't that still required to make the decorator OOT dispatch work?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,139 @@
...
+        x_device = x.device
+
+        # Due to a limitation of the current SpyreModelRunner, 
+        # the input to the SpyreParallelLMHead resides on CPU.
+        # Due to a second limitation regarding sizes that can be used
+        # in a F.linear layer, the original weights need to be padded
+        out = self.maybe_compiled_forward_spyre(
+            convert(x, device=self.weight.device),
+            self.padded_weight.data,
+            bias if bias is not None else None,
```
yannicks1: what about just putting bias here? 😄

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,139 @@
...
+        # an integer.
+        self.padding = 0
+        pad_1 = self.weight.shape[0] % 64
+        if pad_1 == 0:
+            pad_2 = 32 - (self.weight.shape[0] / 64) % 32
+            if pad_2 > 0:
+                self.padding = int(pad_2 * 64)
+                self.padded_weight = F.pad(self.weight, (0, 0, 0, self.padding))
+                logger.warning_once(
+                    "%s: weights padded from %d to %d (torch-spyre limitation)"
```
yannicks1: space missing in two line logger warning sttring

yannicks1 (dismissed):

yannicks1:
do we want a to add a test: `tests/test_parallel_lm_head.py`  similar to `test_rms_norm.py` / `test_silu_and_mul.py` ?

spyre_inference/custom_ops/__init__.py
yannicks1: ah I see it has been addressed on main: https://github.com/torch-spyre/spyre-inference/pull/162

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: Added space for proper text alignment

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: Yes, should work. I adjusted it.

spyre_inference/custom_ops/__init__.py
bohnstingl: Good catch. This was a mistaken. I merged in the latest main branch and this is now fixed.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: Also good catch here. I have adjusted the code to avoid unnecessary padding if the weights already have the correct shape. Also one note here, we only pad the vocabulary dimension of the weights and assume that the hidden dimension of the model has a working shape. This is a basic assumption here, because otherwise any linear layer would fail already before.

bohnstingl: Done.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: Very good catch. This was non-functional as the weights were initialized with random values. I updated the code and now create the padding tensor in the `process_weights_after_loading` function, which is executed after the `load_model` call.

bohnstingl:
@yannicks1 I added a testcase for the `ParallelLMHead`. Currently it also checks the padding procedure, which eventually will be obsolete once torch-spyre is able to handle work division for larger matmuls properly.

spyre_inference/custom_ops/__init__.py
```diff
@@ -17,7 +17,6 @@
 def register_all():
     logger.info("Registering custom ops for spyre_inference")
     vocab_parallel_embedding.register()
-    parallel_lm_head.register()
```
tdoublep, line 18: Why do we remove this?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,143 @@
...
+                f"(quant_config={quant_config}). Only quant_config=None is supported."
+            )
+
+        if self.tp_size > 1:
+            raise NotImplementedError(
+                f"SpyreParallelLMHead does not support Tensor Parallelism "
+                f"(tp_size={self.tp_size}). Only tp_size=1 is supported."
+            )
+
+        logger.debug("Building custom ParallelLMHead for Spyre")
```
tdoublep, line 97: Is this debug log necessary? Do we have them consistently for the other layers?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,143 @@
...
+
+        self.maybe_compiled_forward_spyre = self.maybe_compile(self.forward_spyre)
+
+        if isinstance(self.quant_method, UnquantizedEmbeddingMethod):
+            self.quant_method = SpyreUnquantizedLMHeadMethod()
```
tdoublep: Isn't this check redundant?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,143 @@
...
+
+    def forward_oot(self, x: torch.Tensor, bias: torch.Tensor | None = None) -> torch.Tensor:
+        """OOT forward pass — lm_head matmul on Spyre.
 
-        In the current Spyre eager-mode flow, hidden_states stay on CPU
-        throughout the model. The lm_head weights must also be on CPU
-        for F.linear to work without device mismatches.
+        Called by SpyreUnquantizedLMHeadMethod.apply() from within
+        LogitsProcessor._get_logits(). Converts x to the weight device,
+        runs the compiled F.linear, and converts back to the input device.
```
tdoublep: Aren't the input device and weight device the same?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -2,37 +2,143 @@
...
+            bias: Optional bias tensor
+
+        Returns:
+            Logits tensor [num_tokens, vocab_size] on the input device
         """
-        return self
+        x_device = x.device
+
+        # Due to a limitation of the current SpyreModelRunner, 
+        # the input to the SpyreParallelLMHead resides on CPU.
```
tdoublep: What is this limitation?

spyre_inference/custom_ops/parallel_lm_head.py
yannicks1: maybe we then add an assert on self.weight.shape[1] to catch that? it will currently just fail with a generic backend error, I guess?

spyre_inference/custom_ops/__init__.py
bohnstingl: We don't need this anymore. This was only needed for the opaque torch.compile layer that we used to have in the initial wrapping. With the new ModelRunner, this is not required anymore and a simple import in the `__init__.py` file is enough.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: We do have this debugging statement for other layers as well, but it is not necessary. We can remove all of them and only keep important debugging statements, such as if there is no `dtype` upcast. WDYT?

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: In fact the check shouldn't be done and we should always just set the quantization method to `SpyreUnquantizedLMHeadM`ethod. I've updated.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: No, they are not. The input `x` resides on `cpu`, because before the call to the `ParallelLMHead` and inside the ModelRunner, there are indexing operations being carried out. Those indexing operations require the tensor to be on cpu and that's why they arrive on `cpu`. Only inside the layer, the x gets converted to spyre and the operation is performed on `spyre`. I've updated the comment.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: There are indexing operations taking place. For example https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L4085, https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L4091 or https://github.com/vllm-project/vllm/blob/63ce38a1af43a63a0bed6cd82194b9e1bd3ee4b0/vllm/v1/worker/gpu_model_runner.py#L5608. See also the description of the PR. These operations require the tensor to reside on cpu. Similarly to the above, I've adjusted the comment.

spyre_inference/custom_ops/parallel_lm_head.py
tdoublep: We can do it as a follow-up

tdoublep (approved):
LGTM

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: I'll create an issue, https://github.com/torch-spyre/spyre-inference/issues/172

joerunde:
bot:next-test

bohnstingl:
bot:next-test

bohnstingl:
bot:next-test

joerunde (approved):
lpgtm too!

PR #161, opened 2026-05-04
:bug: fixup the silu and mul tests

joerunde:
#148 was merged without a final test run, and it looks like the `test_siluandmul_oot_dispatch` was broken.

The refactor in 148 removed the `_forward_spyre_impl` method, which this test was mocking to verify the dispatching logic.

This PR removes that check entirely, as there's now no longer an inner method called by `forward_oot` to mock. It should be good enough that we are already validating:
1. our construction of the spyre-specific class
2. _forward_method == forward_oot
3. sillu_and_mul works on spyre


joerunde:
bot:next-test

GOavi101:
LGTM

PR #164, opened 2026-05-05
docs(plugin-test): update plugin test setup with inexact sync flag

prashantgupta24:
Maybe I'm setting something incorrectly but doing `uv sync` without the `inexact` flag removed all dependencies (like `vllm`) for me?


joerunde:
hmm, I think actually just a 
```
uv sync
```
from the top-level should do it since the pytest plugin is part of the workspace and will get installed by default. 

Alternatively, a 
```
uv pip install -e ./tests/plugin
```
would be the way to manually install it

PR #163, opened 2026-05-05
[Docs] Consolidate to include spyre-inference and legacy sendnn

rafvasq:
Consolidates doc build for `spyre-inference` and `sendnn-inference` (legacy) under the existing URL `https://docs.vllm.ai/projects/spyre/`.

- Adds `mkdocs.yaml` and `.readthedocs.yaml` to build the docs site via readthedocs.
- Legacy `sendnn-inference` docs are pulled at build time and served under a `SenDNN` section
- Hooks for generate examples, tables, etc,. are ported from `sendnn-inference` and adapted to handle both the new content and legacy

These preliminary changes must be merged before pointing the RTD project to this repo (currently pointed at sendnn-inference) to start testing.

Works towards #154

PR #169, opened 2026-05-06
🚀 ci: bring up GitHub workflows from sendnn-inference

coderfornow:
Bring up GitHub workflows from sendnn-inference

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,44 @@
+name: 🐛 Bug report
+description: Report a bug in Spyre Inference
+title: "[Bug]: "
+labels: ["bug"]
+
+body:
+- type: markdown
+  attributes:
+    value: |
+      Before submitting, please search [existing issues](https://github.com/torch-spyre/spyre-inference/issues) to avoid duplicates.
+
+- type: textarea
+  attributes:
+    label: Describe the bug
+    description: A clear and concise description of what the bug is.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: How to reproduce
+    description: |
+      Steps to reproduce the behavior. Include a minimal code example if possible.
+      
+    placeholder: |
+      ```
+      # Your code here
+      ```
+      1. Run command '...'
+      2. See error
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: Additional context
+    description: Any other context, logs, etc. of the problem.
+
+- type: checkboxes
+  attributes:
+    label: Checklist
+    options:
+      - label: I have searched for similar issues.
+        required: true
--- a/.github/workflows/check_uv_lock.yml
+++ b/.github/workflows/check_uv_lock.yml
@@ -1,0 +1,48 @@
+name: Check uv.lock
+
+on:
+  pull_request:
+    paths:
+      - 'pyproject.toml'
+      - 'uv.lock'
+  push:
+    branches: [main]
+    paths:
+      - 'pyproject.toml'
+      - 'uv.lock'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+jobs:
+  check-lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: "Get changed files"
+        id: changed-files
+        uses: tj-actions/changed-files@v46
+        with:
+          files_yaml: |
+            root:
+              - pyproject.toml
+              - uv.lock
+
+      - name: "Set up Python 3.12"
+        if: steps.changed-files.outputs.root_any_changed == 'true'
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+          enable-cache: true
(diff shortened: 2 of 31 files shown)


rafvasq (commented):
I think that doc related files (`docs/*`, `.readthedocs`, `mkdocs.yaml`)  can be left out of this PR since it's the focus of #163 and being adapted to handle sendnn_inference and this repo's docs

joerunde:
@coderfornow let's not delete the .md files that already exist under docs, they're correct but just not hosted anywhere yet

.github/workflows/check_uv_lock.yml
```diff
@@ -0,0 +1,48 @@
...
+        id: changed-files
+        uses: tj-actions/changed-files@v46
+        with:
+          files_yaml: |
+            root:
+              - pyproject.toml
+              - uv.lock
+
+      - name: "Set up Python 3.12"
+        if: steps.changed-files.outputs.root_any_changed == 'true' || steps.changed-files.outputs.next_any_changed == 'true'
```
joerunde: `next_any_changed` is dead code, let's remove

tests/plugin/spyre_testing_plugin/sync_upstream_test_deps.py
```diff
@@ -194,15 +194,15 @@ def reorder_dependencies(pyproject_path: Path) -> None:
...
+    new_content = content[: match.start()] + new_section + content[match.end() + 1 :]
 
     with open(pyproject_path, "w") as f:
         f.write(new_content)
 
 
 def main():
     if len(sys.argv) > 1:
-        print(f"Usage: python -m spyre_testing_plugin.sync_upstream_tests", file=sys.stderr)
+        print("Usage: python -m spyre_testing_plugin.sync_upstream_tests", file=sys.stderr)
```
joerunde: ah, there's actually a typo here still. Should be `spyre_testing_plugin.sync_upstream_test_deps`

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: "Set up Python 3.12"
+      uses: astral-sh/setup-uv@v5
+      with:
+        python-version: "3.12"
+        enable-cache: true
+        ignore-nothing-to-cache: true
+        cache-dependency-glob: |
+          pyproject.toml
+    - name: "Install dependencies"
+      run: VLLM_TARGET_DEVICE=empty uv sync --frozen --no-install-package torch-spyre --no-install-package vllm
```
joerunde: `--no-install-package vllm` will clearly fail, since we have to type check against vllm. Does this work with only the `--no-install-package torch-spyre` override?


Otherwise we'll need to edit the `prek` command here to exclude the `ty` check

CLAUDE.md
```diff
@@ -114,4 +121,4 @@ python -m spyre_testing_plugin.sync_upstream_test_deps
 - **Line length**: 100 characters (ruff)
 - **Type hints**: Ignore `possibly-missing-attribute` (ty)
 - **Excluded from ty**: `spyre_inference/__init__.py`
-- **Codespell**: Ignore list includes `dout, te, indicies, subtile, ElementE`
+- **Codespell**: Ignore list includes `doubt, te, indices, subtile, ElementE`
```
joerunde: lol, we typo-corrected the list of known typos!

Should `dout` and `indicies` instead be added to `_typos.toml`?

coderfornow: lol true, missed that

joerunde:
@coderfornow running `./format.sh` on your branch I still see a lot of `ty` failures. Can you at least put inline ignores for each of those with `# TODO:` comments to fix? We don't want to merge this and immediately have broken checks.

Also, I don't think the `VLLM_TARGET_DEVICE=empty` override will work, because pyproject.toml specifies to always use `cpu`. That might cause some long compile times when vllm is updated, but hopefully won't be too much of an issue.

Once you have a passing run of `./format.sh` on your branch, let's go ahead and merge so we can see the checks running. They will probably still have issues, but once they exist on the main branch then we can start fixing them with more PRs before enforcing them

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,46 @@
...
+    description: Any other context, logs, etc. of the problem.
+
+- type: checkboxes
+  attributes:
+    label: Checklist
+    options:
+      - label: I have searched for similar issues.
+        required: true
+
+# Made with Bob
```
bohnstingl: I am not a big fan of those statements, do we need to keep them?

bohnstingl (approved):
By the looks of it, it seems fine to me. I appreciate the effort and highly value having the GH workflows back again 😊. 
I can't really comment much on the technical details, but I would agree that merging this should have priority, as it helps the code quality a lot. If there are issues surfacing, we can correct them as they appear.

coderfornow:
Hey @joerunde , can you review now pipelines are passing with VLLM Target Device CPU :)

pyproject.toml
```diff
@@ -123,6 +123,10 @@ explicit = true
 
 [tool.ty.rules]
 possibly-missing-attribute = "ignore"
+# TODO: re-enable once vllm/torch-spyre imports resolve in CI.
+unresolved-import = "ignore"
+# TODO: re-enable after fixing subclass attribute lookups in custom_ops/linear.py.
+unresolved-attribute = "ignore"
```
joerunde, line 129: This is pretty heavy handed- but we can merge for now and make it better later

joerunde (approved):
Formatting is good to go- let's get this merged and we can fixup the type checking later

PR #171, opened 2026-05-06
[spyre-next]  Remove VocabParallelEmbedding implementation

### LINKED ISSUE #32: [Feature]: Ensure VocabParallelEmbedding layer is compatible with SpyreRunner
### Feature description

The aim of this issue is to rework the SpyreVocabParallelEmbedding layer such that it can consume inputs on `spyre`, create / keep all layer parameters on `spyre` and produce outputs on `spyre`. Ideally all D2H and H2D data transfers should alleviated. 
Furthermore, it is essential to investigate whether the current opaque layer, for example `torch.ops.vllm.spyre_vocab_parallel_embedding`, presenting a barrier for torch.compile can be removed. This would make the entire layer is torch-compileable and offer significant benefits for optimization and efficiency. However, for this to work, the layer implementation on `spyre` needs to work in eager and torch-compile mode, which may not always be true and depends on torch-spyre support.

### Motivation and context

This work is necessary in order to enable a model to run truly E2E on `spyre`, without immediate D2H and H2D transfers.

### Proposed solution

_No response_

### Checklist

R3hankhan123:
Removing the layer wrapping since to due to code simplification forward_oot came to being just 
returning F.embedding which is the the same of the vllm code. After running inference with layer defination written in
vllm the tensors are on spyre and correct output is being generated
Resolves #32 
### Test Plan
Run inference tests with enforce eager as true and false.
### Test Result
Output is correct with both modes and verified that tensors are on spyre


bohnstingl:
As discussed offline, the code after the simplification is so minimal that we can think of removing it entirely

bohnstingl (approved):
LGTM.

@tdoublep @yannicks1 with the ModelRunner in place, this is now the first layer wrapping that can be removed and we purely leverage code from upstream.

tdoublep (approved):
LGTM - this is awesome!!

yannicks1 (approved):
lgtm, love negative diffs 😍

PR #173, opened 2026-05-06
tests: add MLP layer numerical equivalence tests

### LINKED ISSUE #49: Verify numerical equivalence of MLP layer
### Feature description

This issue tracks the progress of ensuring that the MLP layer wrapper for `torch-spyre` is numerically equivalent to the upstream vLLM implementation.

This issue is related to https://github.com/vllm-project/vllm-spyre/issues/736

### Motivation and context

NA

### Proposed solution

_No response_

### Checklist

GOavi101:
-  tests full MLP forward pass (gate_up_proj → SiluAndMul → down_proj) against a reference PyTorch implementation, parametrized over dtypes, token counts, hidden sizes, and intermediate sizes
-  adds tp_group fixture that initializes a real single-rank gloo process group and wires it into vLLM's _TP so linear layers can be instantiated in unit tests without monkey-patching


joerunde:
bot:next-test

joerunde:
bot:next-test
(checking if the out of date branch was the problem)

tests/test_mlp.py
```diff
@@ -0,0 +1,176 @@
...
+        gate_weight,
+        up_weight,
+        down_weight,
+        gate_bias,
+        up_bias,
+        down_bias,
+    )
+    
+    gate_up_out, _ = gate_up_proj.forward(x)
+    assert isinstance(act_fn, SpyreSiluAndMul)
```
joerunde: No need to assert this here, it's covered in the oot registration test

tests/test_mlp.py
```diff
@@ -0,0 +1,176 @@
...
+    activated = act_fn.forward_oot(gate_up_out)
```
joerunde: also no need to use `.forward_oot`, we should just test with `.forward`

tests/test_mlp.py
```diff
@@ -0,0 +1,176 @@
...
+    activated = F.silu(gate_up[..., :d]) * gate_up[..., d:]
+    return F.linear(activated, down_weight, down_bias)
+
+
+@pytest.mark.spyre
+@pytest.mark.mlp
+@pytest.mark.parametrize("num_tokens", [1, 7, 64, 256])
+@pytest.mark.parametrize("hidden_size", [64, 128, 512])
+@pytest.mark.parametrize("intermediate_size", [128, 256, 1024])
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32])
```
joerunde: I know we have float32 for the silu_and_mul tests too, but just curious why we need to include it?

IIUC based on assertions in `platform.py`, we'll only ever load models in fp16 so we won't need to spend the extra cycles here validating fp32

tests/test_mlp.py
```diff
@@ -0,0 +1,176 @@
...
+        RowParallelLinear,
+    )
+    from vllm.model_executor.layers.activation import SiluAndMul
+    from spyre_inference.custom_ops.silu_and_mul import SpyreSiluAndMul
+
+    # Create input
+    torch.manual_seed(0)
+    x = torch.randn(num_tokens, hidden_size, dtype=dtype)
+    
+    # OOT registration replaces these with Spyre versions
```
joerunde: Thoughts on using an actual MLP layer here instead of building the three different layers?

```
  from vllm.model_executor.models.llama import LlamaMLP                                                                                      
                                                                                                                                             
  mlp = LlamaMLP(                                                                                                                            
      hidden_size=hidden_size,                                                                                                                       
      intermediate_size=intermediate_size,                                
      hidden_act="silu",                                                                                                                     
      bias=use_bias,
      disable_tp=True,                                                                                                                       
      prefix="mlp",                                         
  )                                                                                                                                          
```

tests/conftest.py
romitjain, line 1: A lot of conftest like behaviour is also present in this file: https://github.com/torch-spyre/spyre-inference/blob/main/tests/plugin/spyre_testing_plugin/pytest_plugin.py

We should either add these configs there or extract configs out to this file.

cc: @joerunde

tests/test_mlp.py
romitjain, line 1: @GOavi101 @joerunde 

There's a generic way to extract a larger module from a vllm model:

```python
# Extract vLLM model definition
import torch
import torch.nn as nn

from vllm.config import (
    VllmConfig,
    ModelConfig,
    CacheConfig,
    ParallelConfig,
    DeviceConfig,
    set_current_vllm_config,
    CompilationConfig,
)
from vllm.config.load import LoadConfig
from vllm.model_executor.model_loader import get_model
from vllm.distributed.parallel_state import (
    init_distributed_environment,
    initialize_model_parallel,
)

model_path = "ibm-ai-platform/micro-g3.3-8b-instruct-1b"
dtype = torch.float16
device = "cpu"

model_config = ModelConfig(
    model=model_path,
    dtype=dtype,
)

compilation_config = CompilationConfig(
    custom_ops=["all"],
)

vllm_config = VllmConfig(
    model_config=model_config,
    cache_config=CacheConfig(block_size=16),
    parallel_config=ParallelConfig(
        tensor_parallel_size=1,
        pipeline_parallel_size=1,
    ),
    device_config=DeviceConfig(device=device),
    load_config=LoadConfig(load_format="auto"),
    compilation_config=compilation_config,
)

with set_current_vllm_config(vllm_config):
    init_distributed_environment(world_size=1, rank=0, local_rank=0)
    initialize_model_parallel(tensor_model_parallel_size=1)

    model = get_model(vllm_config=vllm_config)

print(model)
```

```
modules = dict(model.named_modules())
layer = modules["model.layers.0.mlp"] # Any model can be extracted from the dict


# Running forward pass through the module
x = torch.randn(1, in_features, dtype=dtype, device=device)
result = layer(x)
```

I feel,  a way to make the above easy to test might be a better path for larger modules (a collection of layers). Individual layers can be tested similarly to what we did earlier.

WDYT

GOavi101:
bot:next-test

tests/plugin/spyre_testing_plugin/pytest_plugin.py
```diff
@@ -659,6 +660,58 @@ def default_vllm_config(monkeypatch):
...
+            store=store,
+            world_size=1,
+            rank=0,
+        )
+        yield
+    finally:
+        if dist.is_initialized():
+            dist.destroy_process_group()
+        if os.path.exists(store_path):
+            os.unlink(store_path)
```
bohnstingl, line 699: This seems to be actually quite similar to what I recently added in https://github.com/GOavi101/spyre-inference/blob/929930820e71815f6a06a5065686d0cd49210dff/tests/test_parallel_lm_head.py#L13-L37. Could we maybe consolidate these two methods? I agree that `pytest_plugin.py` is the right place for this. I didn't notice this when preparing my tests.

tests/test_mlp.py
bohnstingl: I like the approach.

However, although I think having the test comparing a full vLLM MLP with up-projections, activations and down-projections is good, I would think that we should rather test the individual layers in isolation? For example, we currently have three linear layers wrapped:
- SpyreMergedColumnParallelLinear
- SpyreQKVParallelLinear
- SpyreRowParallelLinear

and we should test them individually as well? In the bigger MLP layer we put multiple layers/activations together and we dilute the impact of the individual components?

bohnstingl (commented):
This PR reminds me of the repeating discussion that we having on whether we should use upstream vLLM tests or not? I would actually say that we should aim to reduce the amount of custom tests that we have and try to simply use the upstream tests. For the case of the MLP layers, I see that there are upstream tests, but only in the LoRa section, e.g., https://github.com/vllm-project/vllm/blob/713b28bd0bf4d7f8226ee20940e3ae171fabd09b/tests/lora/test_layers.py#L751. So for the moment, we could keep this test file and expand as needed, but ultimately we should aim to remove it again. WDYT?

@tdoublep what is your take on this as well?

bohnstingl (changes requested):
LGTM, let me trigger the CI tests.

bohnstingl:
bot:next-test

bohnstingl (approved):

GOavi101:
bot:next-test

tests/test_mlp.py
romitjain: @bohnstingl I agree.

For the bigger layers like MLP, there's a parallel effort in getting that tested for HF model definition in `torch-spyre` repo. We can keep that framework for vLLM layers as well, but only for larger layers like MLP, single layer test can still reside here.

@anubhavjana @kmehant Tagging you here so that you are aware of it and can take the discussion forward.

GOavi101:
bot:next-test

joerunde (approved):
lpgtm too!

Issue #154, opened 2026-04-30
Get a docs build up and running


joerunde:
We have some docs already, and we need to be able to write more and publish them somewhere for users.
Currently, the `sendnn-inference` plugin uses readthedocs and publishes to the vllm docs site here: https://docs.vllm.ai/projects/spyre/en/latest/

Three broad approaches to take for the docs for this plugin are:
1. Publish our own docs to gh-pages or other domain that we own. This would be the easiest to do but may lead to churn later if we want to keep docs centralized on the vllm docs site and have to redirect.
2. Spin up a second `readthedocs` project and get it subprojected into vllm to publish to a different url on the vllm docs site, like https://docs.vllm.ai/projects/spyre-next. This is also straightforward, though the naming would be misaligned with the python packages (`sendnn_inference` -> `projects/spyre`, `spyre_inference` -> `projects/spyre-next`)
3. Piggyback on the existing `readthedocs` project and push docs to the same url (https://docs.vllm.ai/projects/spyre) as the legacy plugin, but with a different versioning scheme. This would require some hackery like cloning the docs from this repo inside builds for the sendnn-inference repo, but would be the only way to consolidate the docs for both plugins under a single url (if we want to do that).

cc our documentation champion @rafvasq

rafvasq:
If we want to keep everything under https://docs.vllm.ai/projects/spyre, then think option 3 is the best path. I think we could reverse the "hackery" though, make `spyre-inference` the primary `readthedocs` project, and clone `sendnn-inference` docs into a legacy version/subtree of docs. Then there's no moving things around if `sendnn-inference` is ever retired.

I'm not sure about option 2, maybe it's not a big deal but I think all of the vLLM's subprojects are within their org so I would avoid trying to squeeze another subproject in and use what we already have.

joerunde:
^^ That sounds good to me, do you have the bandwidth to get started on that @rafvasq?

rafvasq:
I'm on it! Will have a pr up soon

rafvasq:
https://torch-spyre.github.io/spyre-inference/latest/

PR #175, opened 2026-05-06
docs: add mkdocs site with gh-pages deployment

rafvasq:
Sets up doc infrastructure for `spyre-inference` using `mkdocs` deployed to GitHub Pages via [mike](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/) for versioned docs. Includes hooks for example and model table generation carried over from `sendnn-inference`

Towards #154


docs/mkdocs/hooks/generate_model_tables.py
```diff
@@ -0,0 +1,194 @@
+"""Generate model configuration tables from model_configs.yaml"""
```
joerunde: Are we good to remove this? Should only be relevant to sendnn-inference, at least for now

docs/examples/offline_inference/torch_spyre_inference.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Torch Spyre Inference
+---
+
+Source <gh-file:examples/offline_inference/torch_spyre_inference.py>.
+
+``````py
+--8<-- "/Users/rafaelvasquez/workspaces/aiu-inference-dev/spyre-inference/examples/offline_inference/torch_spyre_inference.py"
```
joerunde: looks like a generated file- should this be included?

joerunde:
lpgtm, just a tiny bit of cleanup

bohnstingl (commented):
Honestly, I have not too much clue about this documentation part. I can rubber stamp if needed, but from the technical side I cannot contribute anything, I am afraid. 

If you tell me how to build the documentation, or if there is anything else that I can try out, I am happy to do so though.

docs/examples/offline_inference/torch_spyre_inference.md
rafvasq: Ah right, we ignore these [in sendnn-inference](https://github.com/torch-spyre/sendnn-inference/blob/main/.gitignore#L141-L143)

rafvasq:
@bohnstingl Thanks! In this PR I'm also including a [doc building instructions](https://github.com/torch-spyre/spyre-inference/pull/175/commits/e9b15c3eed398e973e1230164463c9be29778eda) just for future reference. There's also a new action "Preview docs" which will build an artifact that a reviewer can download and parse through locally to check changes more easily.

joerunde (approved):
ship it!

PR #177, opened 2026-05-06
:memo: add proper license and copyright

### LINKED ISSUE #176: Setup correct copyright headers and check
We need to setup correct copyright headers in our source files. As an example, see: http://github.com/torch-spyre/torch-spyre/pull/16. We currently have a smattering of `# SPDX-FileCopyrightText: Copyright contributors to the vLLM project`, which is no longer correct after our move out of the vLLM org.

We need a CI check to ensure that all source files have the correct copyright header. Many actions already exist for this, like: https://github.com/viperproject/check-license-header. But it would not be very hard to whip up our own that looks at all .py files in the diff and ensures the correct header exists.

joerunde:
Closes #176 

See https://github.com/apache/skywalking-eyes, this PR uses a tool from the apache org to check that our licenses are correct and exist on every source file.

### DIFF
--- a/.github/workflows/license-header.yml
+++ b/.github/workflows/license-header.yml
@@ -1,0 +1,16 @@
+name: License Header Check
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  check-license-header:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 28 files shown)


joerunde:
The github action won't run on this PR since it's from a fork, but it well begin running on all other PRs once this is merged

tdoublep (approved):
LGTM

PR #186, opened 2026-05-11
:zap: Fixup uv caching for faster pre-commit checks

joerunde:
## Description

Our pre-commit checks are currently _not_ using uv caching, the cache entries are 1KB and every PR is rebuilding vllm from source. This needs to be fixed, but I'm not 100% sure what part of the setup is currently broken. I don't expect that the runs on the PRs themselves will be able to write to the cache because they'll have read-only creds, but this action does run on `main` after the PR is merged and I don't see caches being written back from those builds 🤔 

I don't think we should include `ignore-nothing-to-cache: true`, since we will always build a uv venv and install dependencies, so either we'll pull from cache or install things that should be cached. Experimenting here to see what's going on with that.

Also the `cache-dependency-glob` by default will look for a uv.lock, and we should probably be keying on the lockfile since that can change independently of the pyproject.toml. (Packages can upgrade without changing the supported version ranges).

Update: Looks like a successful cache from this PR! (590MB instead of 1KB)
image

## Test Plan

- [ ] Find out what's broken 

## Checklist

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -24,14 +24,13 @@ jobs:
       with:
         python-version: "3.12"
         enable-cache: true
-        ignore-nothing-to-cache: true
-        cache-dependency-glob: |
-          pyproject.toml
     # Required to build vllm's CPU backend
     - name: "Install system dependencies for vllm CPU build"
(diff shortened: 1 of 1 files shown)


bohnstingl (approved):
Can't really judge what else could be improved, but I am sure this PR helps with the pre-commit checks.

PR #188, opened 2026-05-12
:bug: simply remove vllm from the pre-commit builds for now 😭

joerunde:
## Description

#186 started pushing uv caches for quickly rebuilding our `pre-commit` environment with vllm installed for proper type checking.

However, now other PRs are pulling cache and immediately failing to build vllm because `ninja` is not found. This is a bit puzzling to me because:
1. If vllm is cached, why is it being rebuilt?
2. None of the other build setup changed and vllm wasn't updated so why would ninja now be missing?

The good news is that they're failing in 10 seconds instead of 20 minutes at least.

Debugging this here 😭 

Update: Still not really sure what's going on. But since we've disabled the type checking, we can punt to a future issue to enable it and properly cache vllm builds then. @coderfornow had it right all along with ignoring vllm 😉 

From some input from @tjohnson31415, we'd likely want to use ccache for vllm build caching anyway because the uv cache may be unsafe- it won't take the runtime library versions into account when loading from cache.

## Checklist

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -32,7 +32,7 @@ jobs:
     # torch-spyre cannot install without the spyre runtime libs installed
     # we're not running any tests so we can skip the dev group with `--no-dev`
     - name: "Install dependencies"
-      run: uv sync --frozen --no-install-package torch-spyre --no-dev
+      run: uv sync --frozen --no-install-package torch-spyre --no-dev -v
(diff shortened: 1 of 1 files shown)


joerunde:
Followup issue opened here: https://github.com/torch-spyre/spyre-inference/issues/191

I'm gonna go ahead and just merge this so that we stop bricking all the other PRs and at least keep the formatter running quickly.

PR #189, opened 2026-05-12
docs: improve mike deployment and update README

rafvasq:
## Description

This PR generally improves/fixes a few things about the `mike` deployment and updates the repo's README and doc site landing page.

- fetch gh-pages before mike runs (otherwise probably results in a bug caught in [sendnn-inference/725](https://github.com/torch-spyre/sendnn-inference/pull/725))
- add pip caching to speed up doc previews, correct permissions and concurrent build guards
- add path filters to avoid deploying docs on every PR
- fix the "_you're currently looking at the dev docs_" banner; should now only show on `latest` (`main`) and link to a `stable` release
- Add automated PR comment to link to the doc artifact/provide quick preview instructions for reviewers (only on PRs that touch doc related files)
- Update the repo's README and doc site landing page to include updated links, resources, etc., similar to [sendnn-inference](https://github.com/torch-spyre/sendnn-inference/blob/main/README.md)
- Misc mkdocs improvements

## Test Plan

- Preview docs
- Automated PR comment with doc review instructions won't trigger until merged

## Checklist

- [x] I have updated the documentation (if applicable)

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -16,10 +16,24 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
+      - uses: actions/cache@v4
+        with:
+          path: ~/.cache/pip
+          key: docs-${{ hashFiles('docs/requirements-docs.txt') }}
+          restore-keys: docs-
       - run: pip install -r docs/requirements-docs.txt
       - run: mkdocs build --strict
       - uses: actions/upload-artifact@v4
         with:
           name: preview-docs
           path: site/
+          retention-days: 7
+      - name: Save PR metadata
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: echo "$PR_NUMBER" > pr-number.txt
+      - uses: actions/upload-artifact@v4
+        with:
+          name: pr-number
+          path: pr-number.txt
           retention-days: 7
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -28,6 +28,8 @@ jobs:
           name: preview-docs
           path: site/
           retention-days: 7
+      # The PR number must be passed via artifact because doc_comment.yml runs on
+      # workflow_run (triggered after this workflow), which has no pull_request context.
       - name: Save PR metadata
         env:
           PR_NUMBER: ${{ github.event.pull_request.number }}
--- a/README.md
+++ b/README.md
@@ -1,76 +1,40 @@
-# spyre-inference
+<h1 align="center">
+Spyre Inference
+</h1>
 
-A vLLM platform plugin for IBM Spyre AI accelerators.
+<p align="center">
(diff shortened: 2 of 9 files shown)


.github/workflows/preview-docs.yml
```diff
@@ -16,10 +16,24 @@ jobs:
...
+          key: docs-${{ hashFiles('docs/requirements-docs.txt') }}
+          restore-keys: docs-
       - run: pip install -r docs/requirements-docs.txt
       - run: mkdocs build --strict
       - uses: actions/upload-artifact@v4
         with:
           name: preview-docs
           path: site/
           retention-days: 7
+      - name: Save PR metadata
```
joerunde, line 33: It looks like this pushes a separate zip file with only `pr-number.txt` in it, with the PR number. If this isn't saved along with the docs themselves, then what's it for?

README.md
```diff
@@ -1,76 +1,40 @@
...
-# Install with uv (recommended)
-uv sync --frozen
-```
-
-**Note:** `torch-spyre` compilation requires access to IBM Spyre hardware with the Spyre Runtime stack. See internal development documentation for environment setup.
-
-## Usage
-
-The plugin automatically registers with vLLM when installed.
-Use it by setting `VLLM_PLUGINS=spyre_inference`
```
joerunde, line 33: This important configuration bit I think is now missing entirely from the docs, can we put it back on the installation page, or seed a new configuration.md?

README.md
```diff
@@ -1,76 +1,40 @@
...
-# Clone the repository
-git clone https://github.com/torch-spyre/spyre-inference
-cd spyre-inference
+Visit our [documentation](https://torch-spyre.github.io/spyre-inference/):
 
-# Install with uv (recommended)
-uv sync --frozen
-```
-
-**Note:** `torch-spyre` compilation requires access to IBM Spyre hardware with the Spyre Runtime stack. See internal development documentation for environment setup.
```
joerunde, line 28: This info is on the installation page, but I think there it just says `sendnn` needs to be available but we wanted to say `Spyre Runtime Stack` since sendnn is just one internal component, and it happened to just be removed too.

Can we port this language there?

joerunde (approved):
couple really minor comments, feel free to address as you see fit and get this merged!

.github/workflows/preview-docs.yml
rafvasq: It's used by `doc_comment.yml` which is triggered when this job finishes and runs on the base branch (so it doesn't have PR context). The PR number is passed via artifact so it knows where to put the comment. I added a comment above for context and here's the ([example](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#using-data-from-the-triggering-workflow) I followed). 

I thought that having a comment on a doc PR directing a reviewer how to preview the docs while we're not on RTD is nice to have. I'll see how it works when merged.

README.md
rafvasq: Replaced a couple of `sendnn` references like this one to `Spyre Runtime Stack`

PR #194, opened 2026-05-13
:wrench: use no_grad instead of inference_mode

joerunde:
## Description

What it says on the tin!

I noticed that we get some really odd results when using inference_mode, documented here: https://github.com/torch-spyre/torch-spyre/issues/2029

`no_grad` should be sufficient for us, so until (or if) inference_mode works properly, we should probably prefer the correctness of using no_grad instead.

## Test Plan

- [ ] unit tests

## Checklist


joerunde:
bot:next-test

PR #196, opened 2026-05-15
cicd: add a workflow to run tests on every commit pushed to a PR branch

HarikrishnanBalagopal:
## Checklist

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -102,8 +102,6 @@ jobs:
             --frozen \
             --verbose \
             --all-extras \
-            --active \
-            --inexact \
             --group dev
           # Print the resolved torch install for log visibility.
           uv pip show torch
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -1,0 +1,138 @@
+name: 'test_each_commit'
+
+on:
+  push:
+    branches: [main]
+    tags: ['*']
+  pull_request:
+    branches: [main]
+  merge_group:
+    types: [checks_requested]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+  checks: write
+
+jobs:
+  test:
+    name: 'build and run tests'
+    runs-on:
+      - x86_64
+      - spyre_pf_x1
+      - linux
+      - image_spyre_inference
+    timeout-minutes: 720
+    env:
+      # The CI env var is very important.
+      # GitHub actions automatically sets CI=true
+      # which causes the pytest inductor tests to fail with AssertionError
+      # due to a mix of fallbacks and decompositions.
+      # Here we set it to an empty string (or we could also unset it).
+      CI: ''
+      CLONED_TORCH_SPYRE_DIR: /home/senuser/torch-spyre
+      CLONED_SPYRE_INFERENCE_DIR: /home/senuser/spyre-inference
+
+    steps:
+      - name: Checkout PR branch/commit in pre-cloned spyre-inference
+        env:
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          PR_HEAD_REF: ${{ github.event.pull_request.head.ref }}
+          GITHUB_SHA: ${{ github.sha }}
+          GITHUB_REF: ${{ github.ref }}
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          # The pre-cloned repo may be owned by a different user than the
+          # runner process; mark all directories as safe to avoid git's
+          # "dubious ownership" error.
+          git config --global --add safe.directory '*'
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          # Prefer the PR head SHA when triggered by pull_request, otherwise
+          # fall back to the SHA of the push/tag that triggered the workflow.
+          TARGET_SHA="${PR_HEAD_SHA:-$GITHUB_SHA}"
+          TARGET_REF="${PR_HEAD_REF:-$GITHUB_REF}"
+          echo "TARGET_REF=${TARGET_REF}"
+          echo "TARGET_SHA=${TARGET_SHA}"
+          git fetch --tags --force origin "+${TARGET_SHA}" || \
+            git fetch --tags --force origin
+          git checkout --force "${TARGET_SHA}"
+          git log -1 --oneline
+
+      - name: Gather runner info
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          date
+          pwd
+          whoami
+          id
+          echo "GHA_RUNNER_POD_NODE_NAME $GHA_RUNNER_POD_NODE_NAME"
+          echo "PATH $PATH"
+          echo "HOME $HOME"
+          python3 --version
+          python3 -c 'import sys; print(sys.path)'
+          echo "--- env ---"
+          env | sort
+          echo "--- AIU env ---"
+          env | sort | grep AIU || true
+          echo "--- repo listing ---"
+          ls -lart "$CLONED_SPYRE_INFERENCE_DIR"
+          echo "---"
+
+      - name: Build spyre-inference
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          # uv sync reconciles the .venv against pyproject.toml + uv.lock,
+          # so torch is reinstalled only when the resolved version or source
+          # differs from what's already installed (e.g., a PR bumps the
+          # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
+          # a different index than the pytorch-cpu source in uv.lock).
+          uv add --editable "$CLONED_TORCH_SPYRE_DIR"
+          uv sync \
+            --verbose \
+            --all-extras \
+            --active \
+            --inexact \
+            --group dev
+          # Print the resolved torch install for log visibility.
+          uv pip show torch
+          uv pip freeze
+
+      - name: Run tests
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          echo "Running tests..."
+          uv run pytest -s -vvv
+          echo "tests done"
+
+  # ---------------------------------------------------------------------------
+  # Required check gate -- this job name must match the existing branch
+  # protection rule: "tests / Run Spyre unit tests".
+  # It fans in all parallel jobs so the single required check
+  # passes only when every suite passes.
+  # ---------------------------------------------------------------------------
+  run-spyre-unit-tests:
+    name: run spyre unit tests
+    runs-on: ubuntu-latest
+    needs: test
+    if: always()
+    steps:
+      - name: Check all suites passed
+        run: |-
+          if [[ "${{ needs.test.result }}" != "success" ]]; then
+            echo "One or more test suites failed."
+            exit 1
+          fi
+          echo "All test suites passed."
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -92,11 +92,6 @@ jobs:
           source "$HOME/.bashrc"
           source /etc/profile.d/ibm-aiu-setup.sh
           cd "$CLONED_SPYRE_INFERENCE_DIR"
-          # uv sync reconciles the .venv against pyproject.toml + uv.lock,
-          # so torch is reinstalled only when the resolved version or source
-          # differs from what's already installed (e.g., a PR bumps the
-          # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
-          # a different index than the pytorch-cpu source in uv.lock).
           uv sync \
             --frozen \
             --verbose \
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -97,7 +97,6 @@ jobs:
           # differs from what's already installed (e.g., a PR bumps the
           # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
           # a different index than the pytorch-cpu source in uv.lock).
-          uv add --editable "$CLONED_TORCH_SPYRE_DIR"
           uv sync \
             --frozen \
             --verbose \
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -99,6 +99,7 @@ jobs:
           # a different index than the pytorch-cpu source in uv.lock).
           uv add --editable "$CLONED_TORCH_SPYRE_DIR"
           uv sync \
+            --frozen \
             --verbose \
             --all-extras \
             --active \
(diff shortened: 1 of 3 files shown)


.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,195 @@
...
+      - image_spyre_inference
+    timeout-minutes: 720
+    env:
+      # The CI env var is very important.
+      # GitHub actions automatically sets CI=true
+      # which causes the pytest inductor tests to fail with AssertionError
+      # due to a mix of fallbacks and decompositions.
+      # Here we set it to an empty string (or we could also unset it).
+      CI: ''
+      CLONED_TORCH_SPYRE_DIR: /home/senuser/spyre-inference
```
joerunde: `CLONED_SPYRE_INFERENCE_DIR`?

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
```
joerunde: For now the matrix can probably just include `-m "upstream"` and `-m "not upstream"`. That'll run our own test files separately from the upstream vllm unit tests that we've enable. Both should take about 3 minutes.

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,195 @@
...
+          # fall back to the SHA of the push/tag that triggered the workflow.
+          TARGET_SHA="${PR_HEAD_SHA:-$GITHUB_SHA}"
+          TARGET_REF="${PR_HEAD_REF:-$GITHUB_REF}"
+          echo "TARGET_REF=${TARGET_REF}"
+          echo "TARGET_SHA=${TARGET_SHA}"
+          git fetch --tags --force origin "+${TARGET_SHA}" || \
+            git fetch --tags --force origin
+          git checkout --force "${TARGET_SHA}"
+          git submodule sync --recursive
+          git submodule update --init --recursive --force
```
joerunde: We don't have any submodules, not sure if this will be required

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,195 @@
...
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_TORCH_SPYRE_DIR"
+          source .venv/bin/activate
+          # uv sync reconciles the .venv against pyproject.toml + uv.lock,
+          # so torch is reinstalled only when the resolved version or source
+          # differs from what's already installed (e.g., a PR bumps the
+          # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
+          # a different index than the pytorch-cpu source in uv.lock).
+          # --reinstall-package torch-spyre: always rebuild torch-spyre from
```
joerunde: We shouldn't need a `--reinstall-package spyre-inference` because we're pure-python, there's nothing to rebuild

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,195 @@
...
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_TORCH_SPYRE_DIR"
+          source .venv/bin/activate
+
+          CONFIG="${{ matrix.suite.config }}"
+
+          echo "Running ${{ matrix.suite.name }} tests..."
+          bash tests/run_test.sh "tests/configs/${CONFIG}" -v
```
joerunde: We don't have a test runner script, our command can just be:
```
pytest -m "${matrix.markers}"
```

(plus whatever other pytest flags you want to add).

joerunde (changes requested):
Thanks!

It looks like this is just copy-pasted from torch-spyre though? We'll need to update this to be `spyre-inference` specific

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,143 @@
...
+          echo "--- repo listing ---"
+          ls -lart "$CLONED_SPYRE_INFERENCE_DIR"
+          echo "---"
+
+      - name: Build spyre-inference
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_TORCH_SPYRE_DIR"
+          source .venv/bin/activate
```
joerunde: This doesn't look right?

We don't want to use any environment from `torch-spyre`, to us it's just another dependency that should be resolved and built along with all the others. Why are we doing anything with `CLONED_TORCH_SPYRE_DIR`?

HarikrishnanBalagopal: > We don't want to use any environment from torch-spyre, to us it's just another dependency that should be resolved and built along with all the others. Why are we doing anything with CLONED_TORCH_SPYRE_DIR?

It's just the common `venv` location, we are not rebuilding torch-spyre here

.github/workflows/test_each_commit.yaml
HarikrishnanBalagopal: > We shouldn't need a `--reinstall-package spyre-inference` because we're pure-python, there's nothing to rebuild

It is applicable to pure-python as well

> The flag --reinstall-package in uv allows you to force a re-installation of specific packages, even if they are already present in the environment. This is particularly useful when a package's content has changed without a version bump, such as during local development or when using direct URL dependencies.

HarikrishnanBalagopal:
Update: we have pushed a new spyre-inference image now

> _Currently blocked on Jenkins worker node issues, once we have the spyre-inference image being built again we can get the tests to pass_

HarikrishnanBalagopal:
cc: @kmehant @YashasviChaurasia can you guys please take a look at why the tests are failing?

ashokponkumar:
@kmehant PTAL

kmehant:
@HarikrishnanBalagopal @ashokponkumar working on the fixes, will keep you all posted once ready.

kmehant:
@HarikrishnanBalagopal @ashokponkumar How do you want me to provide the updates needed in this PR? Should I raise a PR to Hari's feature branch? or whichever is comfortable let me know.

HarikrishnanBalagopal:
> @HarikrishnanBalagopal @ashokponkumar How do you want me to provide the updates needed in this PR? Should I raise a PR to Hari's feature branch? or whichever is comfortable let me know.

@kmehant first the image has to get rebuilt by the CI pipeline and then we can retrigger the tests

kmehant:
@HarikrishnanBalagopal https://github.com/torch-spyre/spyre-inference/compare/main...kmehant:spyre-inference:fix-cicd?expand=1


@joerunde There are 2 tests failing due to need for gated model access. How do we want to handle them? Do you want us set a HF_TOKEN secret or xfail them? 
Screenshot 2026-06-01 at 7 14 32 PM

HarikrishnanBalagopal:
@kmehant Thanks! I have pushed your commit into the PR branch

joerunde:
@kmehant We can set an `HF_TOKEN` to download the models, but ideally we should use cached models and avoid downloading any models at test-time. Without caching we've run into pretty frequent failures to pull models from hf hub. Can we set `HF_HOME` to a shared storage location for these runners so that they can all re-use the same hf cache?

joerunde:
@HarikrishnanBalagopal have the fixes been applied to this PR? On the latest `build and run tests` job I still see a compilation failure:

```
      fatal error: numa.h: No such file or directory
          2 |   #include <numa.h>
            |            ^~~~~~~~
      compilation terminated.
```

HarikrishnanBalagopal:
> @HarikrishnanBalagopal have the fixes been applied to this PR? On the latest `build and run tests` job I still see a compilation failure:
> 
> ```
>       fatal error: numa.h: No such file or directory
>           2 |   #include <numa.h>
>             |            ^~~~~~~~
>       compilation terminated.
> ```

no the image just got recently built, I have retriggered the tests

kmehant:
@HarikrishnanBalagopal I think all tests pass except the gated model access part so we are good! 

> We can set an HF_TOKEN to download the models, but ideally we should use cached models and avoid downloading any models at test-time. Without caching we've run into pretty frequent failures to pull models from hf hub. Can we set HF_HOME to a shared storage location for these runners so that they can all re-use the same hf cache?

Regarding this, @HarikrishnanBalagopal do you want to add this to your GHA runners? The model needed is `meta-llama/Meta-Llama-3-8B`.

HarikrishnanBalagopal:
> Regarding this, @HarikrishnanBalagopal do you want to add this to your GHA runners? The model needed is `meta-llama/Meta-Llama-3-8B`.

yeah we can have a runner set that has a pvc mounted for persistent storage, and set `HF_HOME` in the workflow yaml

kmehant:
@joerunde I will raise a separate patch PR with hf_token integration and some coverage on upstream test reuse, for now we have made the meta model test as xfail to let this PR pass through. Is it okay? 

cc: @ashokponkumar @HarikrishnanBalagopal

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,138 @@
...
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          # uv sync reconciles the .venv against pyproject.toml + uv.lock,
+          # so torch is reinstalled only when the resolved version or source
+          # differs from what's already installed (e.g., a PR bumps the
+          # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
+          # a different index than the pytorch-cpu source in uv.lock).
+          uv add --editable "$CLONED_TORCH_SPYRE_DIR"
```
joerunde: I don't think we should be putting any `uv add` commands in this workflow. This changes the pyproject.toml definition, and ends up testing a different configuration than what's in the PR.

There shouldn't be any reason that we need an editable install of torch-spyre, what's the blocker to removing this?

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,138 @@
...
+          uv sync \
```
joerunde, line 95: IMO we need to have the `--frozen` flag here, without it the `uv.lock` file can be updated on the fly, but we must ensure that we're testing exactly the dependencies that are currently in the uv.lock

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,138 @@
...
+            --verbose \
+            --all-extras \
+            --active \
+            --inexact \
```
joerunde: The combo of `--active` and `--inexact` is very dubious, this means that we're potentially relying on the existing state of some pre-baked virtual environment. That's not a great practice for CI, since we need to ensure that our pyproject.toml / uv.lock combo contains the full information required to correctly install the package

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,138 @@
...
+
+      - name: Build spyre-inference
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          # uv sync reconciles the .venv against pyproject.toml + uv.lock,
+          # so torch is reinstalled only when the resolved version or source
+          # differs from what's already installed (e.g., a PR bumps the
+          # torch~=2.10.0 constraint, or the pre-baked venv has a torch from
```
joerunde: We shouldn't be relying on any pre-baked environments, let's instead point `UV_CACHE_DIR` to a shared location for dependency caching.

We may potentially need to add a `uv cache clean torch-spyre` in that case to make sure we always rebuild torch-spyre, since updating the spyre runtime libs won't invalidate the uv cache but would cause torch-spyre to break at runtime since it would have been compiled against older rpms.

HarikrishnanBalagopal: > in that case to make sure we always rebuild torch-spyre,

with the latest change in the PR we simply add `torch-spyre` as an editable dependency before we build spyre-inference

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,138 @@
...
+          uv add --editable "$CLONED_TORCH_SPYRE_DIR"
+          uv sync \
+            --verbose \
+            --all-extras \
+            --active \
+            --inexact \
+            --group dev
+          # Print the resolved torch install for log visibility.
+          uv pip show torch
+          uv pip freeze
```
kmehant, line 102:
```suggestion
      - name: Build spyre-inference
        run: |
          source "$HOME/.bashrc"
          source /etc/profile.d/ibm-aiu-setup.sh
          cd "$CLONED_SPYRE_INFERENCE_DIR"
          uv sync --verbose --group dev
          # Print the resolved torch install for log visibility.
          uv pip show torch
          uv pip freeze
```

.github/workflows/test_each_commit.yaml
kmehant: Hari, lets go with @joerunde's suggestions I have provided a proposed change below that way we are pinning to the locked torch-spyre version while testing on the checked out spyre-inference WDYT?

.github/workflows/test_each_commit.yaml
kmehant: @HarikrishnanBalagopal ^

cc: @joerunde

.github/workflows/test_each_commit.yaml
HarikrishnanBalagopal: is torch-spyre supposed to be pinned at a specific commit before we build spyre-inference? https://github.com/torch-spyre/spyre-inference/blob/02393d47881d185df9b553c206839819aecaf447/pyproject.toml#L96

.github/workflows/test_each_commit.yaml
HarikrishnanBalagopal: Hi, can we put all of this in a Makefile?

```Makefile
.PHONY: build
build:
   uv sync --frozen .........
```

that way we can avoid encoding all of these details only in the CI pipelines

HarikrishnanBalagopal: for now I have added the `--frozen` flag in the workflow itself

.github/workflows/test_each_commit.yaml
HarikrishnanBalagopal: have removed those flags

.github/workflows/test_each_commit.yaml
HarikrishnanBalagopal: @kmehant have made the changes requested

kmehant: thanks @HarikrishnanBalagopal

.github/workflows/test_each_commit.yaml
```diff
@@ -0,0 +1,131 @@
...
+      - name: Build spyre-inference
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          uv sync \
+            --frozen \
+            --verbose \
+            --all-extras \
+            --group dev
```
joerunde, line 99: we don't have any extras, and the `dev` group is included by default by uv (you have to use `--no-dev` or other equivalent flags to exclude is), so these two can be removed as well if you want

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -74,7 +74,7 @@ tests:
     - rel_path: tests/v1/attention/test_attention_backends.py
       allow_list:
         - test: "test_causal_backend_correctness"
-          mode: mandatory_pass
+          mode: xfail
```
joerunde, line 77: it might be nice to toss a comment inline here saying that these are xfail only due to the CI issue to avoid confusion, since these are fairly important tests for our attention backend.

I'm fine with temporarily disabling them, let's make sure we quickly turn around and get the models cached and re-enable them though!

joerunde (approved):
two small followup comments but otherwise lpgtm!

.github/workflows/test_each_commit.yaml
kmehant: @HarikrishnanBalagopal ^

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
kmehant: @joerunde followup is ready, once this is merged, will raise a PR!!

PR #201, opened 2026-05-19
Add Spyre Operation Support Test for Online Softmax

jvlunteren:
## Description

This PR adds a test script to validate Spyre device support for operations required by the online softmax algorithm used in FlashAttention-style fused tiled attention. The script automatically detects unsupported operations (e.g., `torch.maximum`) and implements CPU fallback, enabling full algorithm validation even when certain operations are not yet available on Spyre hardware.

## Related Issues

#121, #147 

## Test Plan

The test plan is integrated into the standalone script, which automatically executes a validation suite including: (1) operation support detection at startup, (2) CPU validation against reference implementation, (3) Spyre execution with hybrid mode, and (4) numerical accuracy verification. Simply run the script to get a report on Spyre's operation support status and algorithm correctness.

## Checklist

- [X] I have added tests for my changes (if applicable)
- [X] I have updated the documentation (if applicable)

### DIFF
--- a/examples/experimental/spyre_online_softmax_check.py
+++ b/examples/experimental/spyre_online_softmax_check.py
@@ -1,3 +1,17 @@
+# Copyright 2026 The Torch-Spyre Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """
 Experimental example for Spyre Inference project.
 This code is for demonstration and testing purposes.
--- a/examples/experimental/spyre_online_softmax_check.py
+++ b/examples/experimental/spyre_online_softmax_check.py
@@ -56,8 +56,10 @@ KNOWN SPYRE BUGS THIS SCRIPT WORKS AROUND
        data when consumed by bmm. .clone() and .contiguous() on the device
        do NOT repair it — they faithfully copy the corrupted source. The
        only known fix is to slice on host, then transfer (see _seq_slice).
-       Symmetric to the destination-side bug already worked around in
-       torch_spyre/ops/eager.py:_is_narrowed_view (spyre__copy_from).
+    3. Seq-dim narrowed-view writes — writing to a sliced Spyre tensor along
+       the seq dimension (e.g. output[:, 32:64, :] = data) corrupts the data.
+       Workaround: transfer output to CPU, perform the write, transfer back.
+       This is the destination-side counterpart to bug #2.
 """
 
 import os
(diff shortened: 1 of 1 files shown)


joerunde:
Nice!

I notice on my machine that this still prints out a lot of warnings about the `aten.argmax.default` falling back to cpu- but I assume we don't care about that because we're only interested in the fact that torch-spyre supports this op, even if it's not run on the cards?

Also there's one warning about a downcast from int64 to int32 during the check for `.max` - I'm not exactly sure what that's referring to though, do you have any idea?

my output:
```
============================================================
Online Softmax — Spyre Operation Support Test
============================================================

Detecting Spyre operation support...
  torch.maximum() — Supported
  torch.bmm() — Supported
  torch.exp() — Supported
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:132: FallbackWarning: aten.argmax.default is falling back to cpu
  a.max(dim=-1, keepdim=True)[0].cpu()
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:132: UserWarning: Backend Spyre does not support int64; downcasting to int32 may change values outside the 32-bit range. You can silence this via warnings.filterwarnings(...) or spyre.set_downcast_warning(False) or TORCH_SPYRE_DOWNCAST_WARN env. variable. (Triggered internally at /home/senuser/.cache/uv/git-v0/checkouts/b06721edd05a1596/c124a56/torch_spyre/csrc/types_mapping.h:96.)
  a.max(dim=-1, keepdim=True)[0].cpu()
  torch.max(dim=-1) — Supported
  torch.sum(dim=-1) — Supported
  arithmetic (mul, add, sub, div) — Supported

Mode: Native Spyre only
============================================================

Creating test inputs:
  Query: [8, 64, 128]
  Key:   [8, 256, 128]
  Value: [8, 256, 128]

Computing reference attention (standard unfused, CPU)...
Computing fused tiled attention (Q_tile=32, KV_tile=64, CPU)...

CPU Validation:
  Max difference:  3.58e-07
  Mean difference: 2.41e-08
  Matches reference!

============================================================
Testing on Spyre Device
============================================================

Creating Spyre tensor layouts and transferring inputs...
  Query: torch.Size([8, 64, 128]), device=spyre:0
  Key:   torch.Size([8, 256, 128]), device=spyre:0
  Value: torch.Size([8, 256, 128]), device=spyre:0

Computing fused tiled attention on Spyre (native Spyre for torch.maximum)...
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:283: FallbackWarning: aten.argmax.default is falling back to cpu
  scores_max = scores.max(dim=-1, keepdim=True)[0]
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:283: FallbackWarning: aten.argmax.default is falling back to cpu
  scores_max = scores.max(dim=-1, keepdim=True)[0]
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:283: FallbackWarning: aten.argmax.default is falling back to cpu
  scores_max = scores.max(dim=-1, keepdim=True)[0]
/home/senuser/spyre-inference/examples/experimental/spyre_online_softmax_check.py:283: FallbackWarning: aten.argmax.default is falling back to cpu
  scores_max = scores.max(dim=-1, keepdim=True)[0]

Spyre Validation:
  Max difference:  3.70e-03
  Mean difference: 2.69e-04
  Matches reference within fp16 tolerance (1e-02)

============================================================
Test Complete
============================================================
Spyre operation support summary:
  Supported: bmm, mul, add, sub, div, exp, sum(dim=-1), max(dim=-1)
  Not supported: (none detected)

  Known bugs requiring workarounds:
    - seq-dim narrowed-view reads on Spyre (bmm yields garbage)
      workaround: slice on host then transfer (_seq_slice)
    - seq-dim narrowed-view writes on Spyre (data corruption)
      workaround: write via CPU round-trip (fused_tiled_attention line 340-344)
```

jvlunteren:
Both warnings are expected and harmless.

- The `aten.argmax.default` CPU fallback: `torch.max(dim=-1)` decomposes into `amax` (Spyre-native, the values) + `argmax` (CPU fallback, the indices). We only use the values,  so the indices fallback is dead work. But yes, we only care that the values path runs on Spyre, which it does.

- The int64 downcast warning comes from the `argmax` return value landing on device. Spyre doesn't support int64 so it silently narrows to int32. Same root cause, same conclusion: indices path we don't use, irrelevant to the test.

joerunde (approved):
lgtm then!

Issue #44, opened 2026-03-12
Build a proof-of-concept for compiling a vector-add in Helion to Spyre


bringlein:
- learn the compiler details of Helion
- understand the lowering of torch-spyre
- figure out a way how to connect these two
- build a proof-of-concept script to demonstrate it 
- if possible: with consideration of Helions autotuner

tdoublep:
Is there a PR open for this? Otherwise please leave In Progress.

bringlein:
no PR, but working example.

bringlein:
state: no update, but since `spyre-inference` is now separated...should I open a PR there under smth like "experimental/helion"? @tdoublep

tdoublep:
@bringlein Yes - let's do it!

bringlein:
https://github.com/torch-spyre/spyre-inference/pull/203

tdoublep:
@tdoublep to review this week

tdoublep:
Monthly cleanup: this issue carried the `May2026` label and is rolling into `June2026`. @bringlein — could you share a quick status update and a new estimated completion date? If the work has been deprioritized or blocked, please note that here too so anyone tracking this knows what to expect.

bringlein:
thanks @tdoublep (or the bot writing on his behalf...) the last status is it is waiting for a review from @tdoublep since May 20 😉

PR #203, opened 2026-05-20
Experimental Helion-to-spyre compiler

bringlein:
## Description

as discussed here https://github.com/torch-spyre/spyre-inference/issues/44 

## Checklist


joerunde:
@bringlein just trying to run this with a uv environment built off the current main branch and I see that `torch_spyre._inductor.passes.core_division_planning` doesn't exist. (And I can't compile the older version of torch-spyre still pinned in this branch since all the base rpms I have installed in my dev environment are no longer compatible)

Not sure if that was removed or refactored to a different spot, but at least it doesn't seem to work out of the box any longer

Issue #135, opened 2026-04-27
Add tp>1 support for Embed Layer


dilipgb:
Add vocabulary sharding and TP masking to SpyreVocabParallelEmbedding

Issue #134, opened 2026-04-27
Add tp>1 support for Linear Layers


dilipgb:
Update SpyreMergedColumnParallelLinear and SpyreRowParallelLinear with weight sharding and communication primitives

R3hankhan123:
Looks like numerical variance is there
```
STDERR:
[rank1]: Traceback (most recent call last):
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 324, in <module>
[rank1]:     main()
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 318, in main
[rank1]:     PROBES[args.probe](device, device_group, world_size, rank)
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 250, in probe_row_parallel_linear
[rank1]:     torch.testing.assert_close(out, expected, atol=5e-3, rtol=5e-3)
[rank1]:   File "/home/rehankhan/dt-inductor/.venv/lib64/python3.12/site-packages/torch/testing/_comparison.py", line 1600, in assert_close
[rank1]:     raise error_metas[0].to_error(msg)
[rank1]: AssertionError: Tensor-likes are not close!
[rank1]: Mismatched elements: 3 / 8192 (0.0%)
[rank1]: Greatest absolute difference: 0.00830078125 at index (9, 293) (up to 0.001 allowed)
[rank1]: Greatest relative difference: 0.03704833984375 at index (6, 402) (up to 0.001 allowed)
```

R3hankhan123:
Numerical variance is present in TP=1 also

PR #204, opened 2026-05-20
✨  Add initial distributed device setup

### LINKED ISSUE #137: Implement distributed process groups in worker to support tp>1
Set up distributed process groups and implement TorchSpyreModelRunner

joerunde:
## Description

Initial TP support: lands the distributed-init half (process-group setup, platform overrides, Spyre-aware DeviceCommunicator) so a TP=2 worker brings up `init_process_group` + spyreccl + a working forward-path
`all_reduce`.

Does **not** yet land working TP>1 end-to-end forward — that needs the TP-aware custom layers (#134, #135) and an `all_gather` fallback.

## Changes

- **torch-spyre bump** to pick up the `spyreccl` distributed backend.
- **Platform** (`spyre_inference/platform.py`):
    - `dist_backend = "cpu:gloo,spyre:spyreccl"` so `cpu_group` (gloo) and
      `device_group` (spyreccl) coexist.
    - `get_device_communicator_cls() -> SpyreCommunicator`.
    - Reject `data_parallel_size > 1`; DP+TP not validated yet.
- **Worker** (`spyre_inference/v1/worker/spyre_worker.py`):
    - Pin each worker to its card via `torch.spyre.set_device` before
      spyreccl init.
    - Populate `RANK` / `WORLD_SIZE` / `LOCAL_RANK` / `LOCAL_WORLD_SIZE`
      before `init_process_group`. libspyre_comms reads these via
      `getenv` and throws if unset; vllm's `MultiprocExecutor` uses
      TCP-based init and doesn't set them.
- **SpyreCommunicator** (`spyre_inference/distributed/spyre_communicator.py`):
    - TP=2 manual `all_reduce` via send/recv + broadcast (the only
      collectives implemented in the installed libspyre_comms). TP>2
      raises.
    - `all_gather`, `gather`, `reduce_scatter` raise NotImplementedError
      describing what's needed to unblock each.

## Related Issues

Closes #137 

## Test Plan

- `tests/test_distributed_tp2.py` (new):
    - `test_tp2_tensor_model_parallel_all_reduce` — passes today.
      Subprocess + env-rendezvous + real `init_worker_distributed_environment`
      + manual `all_reduce` probe.
    - `test_tp2_llm_construction` — xfail-strict. `LLM(tp=2)` end-to-end;
      flips when #134/#135 land.
    - `test_tp2_llm_generate_matches_tp1` — xfail-strict. Greedy-decodes
      opt-125m at TP=1 and TP=2, asserts first 2 tokens match per prompt
      (later divergence expected from float16 reduction-order).
- `tests/test_spyre_comms_native_probes.py` (new): four xfail-strict
  probes (`all_reduce`, `all_gather_into_tensor`, `all_gather` list,
  `gather`) on a real spyreccl device_group. When libspyre_comms gains
  a native impl and a comms RPM lands, the corresponding probe flips to
  passing → strict-xfail fails CI → signal to delete the matching
  fallback in SpyreCommunicator.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ extra-build-variables = { vllm = { VLLM_TARGET_DEVICE = "cpu" } }
 # NB: torch-spyre can only be compiled where `sendnn` is available
 [tool.uv.sources]
 vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.20.1" }
-torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "539a1a0c2d54b228c19c982ee80044e210bcde98" }
+torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "c124a56ed06b3d31099f825c5ffaa350007d6809" }
 torch = [
   { index = "pytorch-cpu" },
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ extra-build-variables = { vllm = { VLLM_TARGET_DEVICE = "cpu" } }
 # NB: torch-spyre can only be compiled where `sendnn` is available
 [tool.uv.sources]
 vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.20.1" }
-torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "c124a56ed06b3d31099f825c5ffaa350007d6809" }
+torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "ee37a9de5aaa33995fcbeac36ca1b1131677ce6f" }
 torch = [
   { index = "pytorch-cpu" },
 ]
--- a/spyre_inference/distributed/spyre_communicator.py
+++ b/spyre_inference/distributed/spyre_communicator.py
@@ -1,0 +1,190 @@
+# Copyright 2026 The Spyre-Inference Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""DeviceCommunicator override for IBM Spyre devices.
+
+The installed `libspyre_comms.so` (1.0.0-0.main.1+27.5f812da on this pod)
+has only `barrier`, `broadcast`, single-tensor `allgather(Tensor&,Tensor&)`,
+and pairwise `send`/`recv` actually implemented. The list-form `allgather`,
+`allreduce`, `gather`, and `reduce` are throw-stubs in
+`spyre_comms_internal::SpyreCommsContext`. The `_allgather_base` entry point
+on the torch-spyre side is also stubbed regardless of the comms lib.
+
+That means the base `DeviceCommunicatorBase` collectives that route through
+`dist.all_reduce`, `dist.all_gather_into_tensor`, `dist.all_gather` (list),
+and `dist.gather` will all throw at runtime when the device_group's backend
+is `spyreccl`.
+
+This class supplies a hand-rolled fallback for `all_reduce` that works for
+TP=2 by using send/recv + broadcast (all of which ARE implemented), so the
+TP forward path can run end-to-end on two ranks without waiting for the
+upstream comms-side implementations to land. Other collectives raise a
+clear NotImplementedError that names the upstream PR they're blocked on.
+
+REPLACE-WITH-NATIVE markers below identify each fallback that should be
+removed once the corresponding upstream change lands. The intent is that
+when the comms library catches up, this whole file can be deleted (or
+reduced to a couple of overrides for ops we genuinely want to do
+differently for perf), and the platform's
+`get_device_communicator_cls` can revert to the base class.
+
+Upstream tracking:
+  - all_reduce        : github.ibm.com/ai-chip-toolchain/spyre-comms PR #97
+                        (dsolt/allreduce_work). Currently *draft* and
+                        conflicting with main.
+  - gather            : github.ibm.com/ai-chip-toolchain/spyre-comms PR #101
+                        (dsolt/concat_based_gather). Open, non-draft, awaiting
+                        review + Jenkins CI.
+  - all_gather (list) : no PR identified at time of writing. The single-
+                        tensor `allgather(Tensor&,Tensor&)` IS implemented
+                        in libspyre_comms.so but is not exposed via
+                        torch-spyre's spyre_ccl backend.
+  - reduce            : no PR. `src/coll/reduce.cpp` doesn't exist either,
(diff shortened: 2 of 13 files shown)


joerunde:
bot:next-test

joerunde:
NB: this will require a base image update to pick up new spyre_comms changes. Testing that above

joerunde:
Looks like we have an unrelated regression with MHA failing- documented here: https://github.com/torch-spyre/torch-spyre/issues/2195

joerunde:
bot:next-test

pyproject.toml
```diff
@@ -93,7 +93,7 @@ extra-build-variables = { vllm = { VLLM_TARGET_DEVICE = "cpu" } }
 # NB: torch-spyre can only be compiled where `sendnn` is available
 [tool.uv.sources]
 vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.20.1" }
-torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "539a1a0c2d54b228c19c982ee80044e210bcde98" }
+torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "c124a56ed06b3d31099f825c5ffaa350007d6809" }
```
prashantgupta24: for some reason I can't seem to build with this torch-spyre version on my dev pod - 

```
    Updated https://github.com/torch-spyre/torch-spyre (c124a56ed06b3d31099f825c5ffaa350007d6809)
  × Failed to build `torch-spyre @ git+https://github.com/torch-spyre/torch-spyre@c124a56ed06b3d31099f825c5ffaa350007d6809`
```

(long error)
```
      /opt/ibm/spyre/deeptools/include/util/print_utils.h:38:13: warning: ‘std::string PrintUtil::printVecwithQuotes(const std::vector<T>&) [with Dtype = long int]’ defined but not used [-Wunused-function]
         38 | std::string printVecwithQuotes<int64_t>(const std::vector<int64_t>& myVec) {
            |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      ninja: build stopped: subcommand failed.
```

prashantgupta24: I installed from main first (uv sync --frozen) which installed `https://github.com/torch-spyre/torch-spyre?rev=539a1a0c2d54b228c19c982ee80044e210bcde98`. Then I copied over code from this branch and tried to run uv sync again, which failed with the above error

joerunde: yup- this requires later builds of the runtime components which are currently in `dev-next-test` which is where we stage updates to the `dev-next` image

tests/test_spyre_attn.py
```diff
@@ -163,7 +163,13 @@ def ref_attn(
     "num_heads",
     [
         pytest.param((32, 8), id="GQA"),
-        pytest.param((32, 32), id="MHA"),
+        pytest.param(
+            (32, 32),
+            id="MHA",
+            marks=pytest.mark.skip(
+                "Most MHA cases fail to compile, see https://github.com/torch-spyre/torch-spyre/issues/2195"
```
tdoublep: Do we need issue to track this on our side? Just so we don't forget to re-enable it once this is resolved.

tests/test_spyre_attn.py
```diff
@@ -208,13 +214,13 @@ def test_spyre_attn(
     num_blocks: int,
 ) -> None:
     """Validate SpyreAttentionImpl against a reference implementation."""
+    num_query_heads, num_kv_heads = num_heads
```
tdoublep, line 211: nit: why do we need to move this?

tests/test_distributed_tp2.py
```diff
@@ -0,0 +1,158 @@
...
+        return min(len(a), len(b))
+
+    tp1 = run(1)
+    tp2 = run(2)
+    for i, (a, b) in enumerate(zip(tp1, tp2)):
+        n = _matching_prefix_len(a, b)
+        assert n >= 2, (
+            f"prompt {i}: tp1 and tp2 diverged at token {n} "
+            f"(expected >=2 matching tokens). tp1={a} tp2={b}"
+        )
```
tdoublep, line 158: nit: is there any existing util for doing this kind of comparison (either upstream or otherwise)?

tdoublep (approved):
This is awesome!! Couple small questions but nothing blocking merge imo

tests/test_distributed_tp2.py
```diff
@@ -0,0 +1,158 @@
...
+        return result
+
+    def _matching_prefix_len(a: list[int], b: list[int]) -> int:
+        for i, (x, y) in enumerate(zip(a, b)):
+            if x != y:
+                return i
+        return min(len(a), len(b))
+
+    tp1 = run(1)
+    tp2 = run(2)
```
prashantgupta24: nit, but can we add it like

```suggestion
    tp1 = run(tp=1)
    tp2 = run(tp=2)
```

I was confused for a while

spyre_inference/distributed/spyre_communicator.py
```diff
@@ -0,0 +1,196 @@
...
+        what's needed to unblock them.
+    """
+
+    def all_reduce(self, input_: torch.Tensor) -> torch.Tensor:
+        # World size 1 is a no-op for all_reduce; the base class doesn't
+        # short-circuit here, so do it ourselves.
+        if self.world_size == 1:
+            return input_
+
+        if self.world_size != 2:
```
prashantgupta24, line 111: For TP>2, what's the correct send/recv pattern? Do we care to try and implement that?

joerunde: I think the correct thing is to wait for real collectives- this was just a bandaid to be able to test TP=2.

Gather did just merge in spyre-comms so I'll check that today

joerunde: update: this was the single-tensor form of gather that merged in spyre-comms (the output looks like torch.concat) but torch-spyre still wires up the stub list form of gather (where the result is a vector of tensors). 

The high level epic for this seems to be here: https://github.com/torch-spyre/torch-spyre/issues/1902

tests/test_spyre_attn.py
joerunde: oh I think there was a reverted change here that did an assertion about these to check for MHA cases and skip them up front, but then the skip got refactored to the marker

tests/test_spyre_attn.py
joerunde: Comment on that issue says it's already fixed so I'll go ahead and check that too

tests/test_distributed_tp2.py
joerunde: yeah ideally we'd want something stronger like what we've been doing in sendnn-inference with golden token injection and validating the full logits tensor over the sequence. We could pull that in, but I think we wanted to push better testing strategies upstream instead of vendoring them here.

Last I checked the upstream tests only look at the first token and check that it matches :/. I'll have claude take another gander

tests/test_distributed_tp2.py
joerunde: sure!

tests/test_spyre_attn.py
joerunde: looks like it's good to go!

joerunde:
bot:next-test

joerunde:
Thanks for the reviews @tdoublep and @prashantgupta24!

The status after the torch bump is that there's one import-time side effect to work around. Our good friend claude whipped up a few candidate fixes but I'll have to get around to implementing one of them tomorrow, then this should be good to merge.

joerunde:
bot:next-test

validating the spyre autoload fix

PR #206, opened 2026-05-21
first implementation of list-based paged_attn

### LINKED ISSUE #205: Implement list-based paged attention
As next step to test true paged attention, we want to implement a backend that manages the KV-cache as cpu-side lists, and calls online softmax for each individual page. 
See #80 #5 #147 etc...

bringlein:
## Related Issues
#205 

## Checklist


tests/test_spyre_attn.py
```diff
@@ -24,26 +25,39 @@
...
-        test_tensor = torch.randn(1, device=torch.device("spyre"))
-        del test_tensor
+        torch.randn(1, device=torch.device("spyre"))
+        return True
     except Exception:
-        pytest.skip("Spyre device not available - these tests require Spyre hardware")
+        return False
+
+
+SPYRE_AVAILABLE = _spyre_available()
```
joerunde: See the deleted comment above:
```
    """Lazy check that spyre devices are available.
    This must be done lazily to avoid accessing a device at import time.
    """
```

That could have been clearer- the reason this must be done lazily is that as soon as the main pytest process accesses a spyre card, it becomes unavailable for any subprocess to use. That means that all tests that use e.g. `vllm.LLM` will fail, because they need to access the spyre cards from within subprocesses. Our pytest plugin already organizes the tests so that all the tests using subprocesses run first.

This should be called from within the `configure_device` fixture to avoid the import-time side effect

tests/test_spyre_attn_cpu.py
```diff
@@ -0,0 +1,281 @@
...
+    passed = max_diff < atol
+
+    status = "PASS" if passed else "FAIL"
+    seqs_str = ",".join(f"q={q}/kv={kv}" for q, kv in seq_lens_list)
+    label = f"[{seqs_str}] heads={num_query_heads}/{num_kv_heads},d={head_size}"
+    print(f"  [{status}] {label}  max_diff={max_diff:.4e} mean_diff={mean_diff:.4e}")
+    return passed
+
+
+if __name__ == "__main__":
```
joerunde: questions:
1. Shouldn't this be a `pytest` test? If it's intended to be used as a test, it needs to be discoverable and runnable by our test infrastructure
2. what in here isn't covered by `test_spyre_attn`? It looks like that's already had some handling added to it to be able to run on cpu

tests/test_spyre_attn.py
```diff
@@ -24,26 +25,39 @@
...
+def configure_device(monkeypatch):
+    """Configure overwrite_f based on hardware availability.
+
+    On Spyre: enable overwrite_f (uses compiled path).
+    On CPU: disable overwrite_f (uses eager slice assignment).
+    """
+    if SPYRE_AVAILABLE:
+        monkeypatch.setenv("SPYRE_USE_OVERWRITE_F", "1")
+    else:
+        monkeypatch.setenv("SPYRE_USE_OVERWRITE_F", "0")
```
joerunde: IMO changing the test input based on environment is an antipattern, this should instead be handled by parameterizing the test and skipping the cases for spyre when it's not available. That way the test reports are consistent, and no matter what environment the tests are run in we get a clear signal about what ran so we know exactly what passes and what fails

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -88,37 +197,163 @@ def __init__(
...
+            num_blocks_s = (kv_len_s + block_size - 1) // block_size
+            for b in range(num_blocks_s):
+                col_start = b * block_size
+                col_end = col_start + block_size
+                tile = mask_4d_cpu[row_start:row_end, :, :padded_query_len, col_start:col_end]
+                seq_tiles.append(tile.contiguous().to(self._target_device))
+            attention_mask_tiles.append(seq_tiles)
+
+        # Full mask on device (kept for backward compat, may be removed later)
+        attention_mask = mask_4d_cpu.to(device=self._target_device)
```
joerunde: are there any usages of `attention_metadata.attention_mask`? I thought that was only used by the experimental attention backend in `spyre_attn_exp.py`

Speaking of - do these attention changes now overcome `spyre_attn_exp`?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,448 +434,241 @@ def __init__(
...
+            # Reshape back: [num_kv_heads, num_queries_per_kv, padded_query_len, head_size]
+            #   → [query_len, num_heads, head_size]
+            result = result.reshape(1, num_heads, padded_query_len, head_size)
+            result = result.transpose(1, 2).contiguous()
+            seq_result = result[0, :query_len, :, :]
+
+            # Write into output buffer
+            if use_overwrite_f:
+                seq_result_dev = seq_result.contiguous().to(self._target_device)
+                output = torch.ops.spyre.overwrite_f(
```
joerunde: I forget the semantics of `overwrite_f`- will this assignment to `output` here assign the local to a new tensor instead of doing the update in-place?

joerunde:
I understand this is a draft- just left some comments mostly on the test setup that we should consider before this is ready for a thorough review

tests/test_spyre_attn.py
bringlein: thanks, fixed.

tests/test_spyre_attn.py
bringlein: I agree, I didn't check the spyre part of the tests thoroughly yet.

bringlein:
thanks for the early feedback @joerunde, it is helpful! (and sorry for the noise earlier, that was me clicking on the wrong approve button :D :see_no_evil: ).

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: yes and no. in eager mode, the functional `overwrite_f` returns a new output. If using torch.compile, the compiler converts it to an in-place op....at least, AFAIK.

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: good catch, that had to be cleaned up. 
> Speaking of - do these attention changes now overcome spyre_attn_exp?

likely, yes.

tests/test_spyre_attn_cpu.py
bringlein: I wanted to get a quick cpu only test...and hence an extra script seemed easier. I removed it now.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -39,43 +51,144 @@
...
+
+    def fn(q, k_pages, v_pages, page_indices, mask_tiles, scale):
+        tile_max = None
+        tile_sum = None
+        tile_output = None
+
+        for i in range(num_blocks):
+            page_idx = page_indices[i]
+            k_page = k_pages[page_idx]
+            v_page = v_pages[page_idx]
```
bringlein: @dgrove-oss @shruthip-venkatesh @pradghos @joyalbin : these are the two lines where indirect access happens (n this example, in a future example `q` would also be indirect access). Just in this case `k_pages` and `v_pages` are lists of tensor objects on the CPU instead of tensors on the device.

bringlein: further comments on indirect access see https://github.com/torch-spyre/torch-spyre/issues/866

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -39,43 +47,202 @@
...
+        torch.ops.spyre.overwrite(input, output, dims, offsets)
+    else:
+        # intended behaviour on cpu
+        sliced_t = output
+        for i, dim in enumerate(dims):
+            sliced_t = torch.narrow(sliced_t, dim, offsets[i], 1)
+        sliced_t.copy_(input)
+
+
+def _indirect_matmul_mock(
```
bringlein, line 73: I implemented here an example of a potential custom op `_indirect_matmul` if we can't get views to work with indirect access.

bringlein:
ok, so the current state
1. unit test passes on CPU and Spyre, but 
    - I had to increase the numerical error...which I don't know if it is a problem (it always only affected like %0.1 percent of the outputs) 
    - I had to increase the torch.compile recompile limit to 1024 (instead of 256)
 2. end2end also runs...but return garbage output: 
 
```
python3 examples/offline_inference/torch_spyre_inference.py --model ../../models/ibm-granite/granite-3.3-8b-instruct/ --max-model-len 1024 --max-num-seqs 1 --num-gpu-blocks-override 64 --num-prompts 1 --attention-backend CUSTOM
....
===============
CompletionOutput(index=0, text='0. , The,,,,,,ified ( /rand, Dan,\n0', token_ids=[34, 32, 225, 30, 886, 30, 30, 30, 30, 30, 30, 1639, 308, 517, 6386, 30, 22541, 30, 203, 34], routed_experts=None, cumulative_logprob=None, logprobs=None, finish_reason=length, stop_reason=None)
===============

Prompt:
 'What are IBMs main businesses?'

Generated text:
 '0. , The,,,,,,ified ( /rand, Dan,\n0'

```

I guess this could be related to a) the numerical issues in unit tests (although it was always like 32/131072 elements had a difference of 0.45 instead of 0.3 allowed...so I don't know if that is an issue) or b) the k/v caches are not populated correctly? 
I would appreciate some help & ideas @jvlunteren @bohnstingl ...thanks!

bringlein:
Also, the current implementation does not work for MHA or MLA models...since then the individual k-pages cannot be re-stickified during the online softmax calculation. I tried to fix it, but there is no trivial solution like reshaping etc. specifically for MHA only. So this is also a TODO.

bohnstingl:
I pushed some fixes to make the output correct in E2E tests, @bringlein could you please verify that?

A main issue that we are struggling with in this implementation is the fact that any slicing operation needs to take place on `CPU`, as it is not yet supported on `spyre`. We have to work around this issue by moving the respective tensor to CPU before doing the slicing and then moving it back. Unfortunately, the slicing results in a silent correctness error. So it only surfaces as wrong outputs, rather than as a hard error that the developer can see.
There are already several issues open on the `torch-spyre` side to track this, see for example https://github.com/torch-spyre/torch-spyre/issues/268.

One additional way to reduce the impact of the slicing on the attention could be to separate decode and prefill batches, as this would allow to save some slicings. However, this feels to me like a big sidestep, without major benefits.

cc @tdoublep

bringlein:
thanks for your help @bohnstingl , I can confirm it works end2end now also in my test-setup

```
python3 examples/offline_inference/torch_spyre_inference.py --model ibm-granite/granite-3.3-8b-instruct/ --max-model-len 1024 --max-num-seqs 1 --num-gpu-blocks-override 64 --num-prompts 1 --attention-backend CUSTOM

....


CompletionOutput(index=0, text="\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,", token_ids=[203, 203, 36671, 1182, 2575, 6866, 46654, 2305, 6564, 21909, 11258, 30, 6564, 17265, 11258, 30, 3859, 30, 18919, 30], routed_experts=None, cumulative_logprob=None, logprobs=None, finish_reason=length, stop_reason=None)
===============

Prompt:
 'What are IBMs main businesses?'

Generated text:
 "\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,"

-----------------------------------

```

but unit tests are failing now?

examples/offline_inference/torch_spyre_inference.py
```diff
@@ -13,24 +13,32 @@
...
-plugin code. So far the new stack (torch-spyre) is simply using upstream vLLM CPU
-worker/runner classes.
+This example shows how to run offline inference on Spyre using the torch-spyre
+plugin code with the TorchSpyreModelRunner.
 
 Optionally, individual layers can be offloaded to Spyre via --custom_ops:
-  - "all": Run all supported ops on Spyre (default)
-  - "none": Run entirely on CPU
+  - "all": Run all supported ops on Spyre (default when enforce_eager=False)
+  - "none": Run entirely on CPU (default when enforce_eager=True)
```
tdoublep: Does this --custom_ops stuff even make sense now that we are using the TorchSpyreModelRunner? Previously, when we were using teh CpuModeRunner, if we disabled a specific CustomOp that would indeed ensure that it runs on CPU, but I believe this is no longer the case. Disabling a specific CustomOp will just mean we try to run the vLLM native implementation on Spyre which may or may not work. Should we consider removing this option from the script?

examples/offline_inference/torch_spyre_inference.py
```diff
@@ -13,24 +13,32 @@
...
   - "+LayerName": Selectively enable specific layers on Spyre
     (e.g., --custom_ops +RMSNorm +SiluAndMul)
 
-Use --enforce_eager to skip torch.compile and run in eager mode.
+Use --enforce-eager to skip torch.compile and run in eager mode.
+
+Environment variables must be set BEFORE importing vLLM:
+  - VLLM_TARGET_DEVICE=spyre_inference: Selects the Spyre platform
```
tdoublep: This is a build-time environment variable and does not need to be set at runtime.

examples/offline_inference/torch_spyre_inference.py
```diff
@@ -13,24 +13,32 @@
...
+  - VLLM_PLUGINS=spyre_inference,spyre_inference_ops: Loads the plugins
 """
 
+import os
+
+# Set environment variables BEFORE importing vLLM
+os.environ["VLLM_TARGET_DEVICE"] = "spyre_inference"
+os.environ["VLLM_PLUGINS"] = "spyre_inference,spyre_inference_ops"
```
tdoublep: I thought that vLLM will automatically detect which plugins are installed?

examples/offline_inference/torch_spyre_inference.py
```diff
@@ -62,9 +70,9 @@ def parse_args():
     parser.add_argument(
         "--attention-backend",
         type=str,
-        default=None,
+        default="CUSTOM",
         dest="attention_backend",
-        help="Attention backend to use (e.g., SPYRE, CPU)",
+        help="Attention backend to use (e.g., CUSTOM for Spyre, CPU)",
```
tdoublep: What is the point of this `--attention-backend` argument? What will happen if it is set to "CPU" - will it even work? If I understand correctly we switch between the two Spyre backends using an env variable. So we seem to have two orthogonal mechanisms for changing the attention backend?

examples/offline_inference/torch_spyre_inference.py
```diff
@@ -154,6 +162,16 @@ def main():
...
     # Create an LLM.
+    # The spyre_inference platform is loaded via VLLM_PLUGINS.
+    # Set VLLM_TARGET_DEVICE to select the platform.
+    os.environ["VLLM_TARGET_DEVICE"] = "spyre_inference"
+    os.environ["VLLM_PLUGINS"] = "spyre_inference,spyre_inference_ops"
+    print(
+        f"[torch_spyre_inference] VLLM_TARGET_DEVICE={os.environ['VLLM_TARGET_DEVICE']!r} "
+        f"VLLM_PLUGINS={os.environ.get('VLLM_PLUGINS')!r}",
+        flush=True,
+    )
```
tdoublep: See comments above. I don't think these are necessary, but even if they are, why do we set them again?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -12,14 +12,21 @@
...
-This backend aims to implement attention using only PyTorch native operations,
-such as matmul, softmax, etc. It supports vLLM's KV cache.
+This backend implements attention using individual page tensors (one per KV block)
+and FlashAttention-style online softmax that iterates over pages. Each page is a
+complete tensor [num_kv_heads, block_size, head_size] on the target device, passed
+directly to bmm without any slicing.
+
+Architecture:
+  - KV cache: two separate lists (k_pages, v_pages), managed by the model runner.
+  - reshape_and_cache: writes new K/V tokens into pages via _overwrite (all on device).
```
tdoublep: One of the problems with having verbose comments like this is that they can get out-dated very quickly. I think even within this PR the reshape_and_cache is not happening all on device, we need to move the K,V to CPU to be able to slice them. Do we really need to have these architectural notes as comments like this?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -51,27 +58,228 @@
...
+    config = get_current_vllm_config_or_none()
+    if config is None:
+        return False
+    cfg = config.compilation_config
+    if cfg.mode == CompilationMode.NONE:
+        return False
+    return cfg.backend != "eager"
+
+
+_ENABLE_COMPILE = _resolve_compilation_config()
```
tdoublep: It seems kinda weird to do this in this specific file? Isn't querying the compilation config something we need to do across the codebase? How do we do this e.g., in the custom ops?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -51,27 +58,228 @@
...
+def _maybe_compile(fn):
+    """Compile fn unless compilation is disabled.
+
+    Mirrors the gating in CustomOp.maybe_compile without requiring CustomOp
+    inheritance. Skips torch.compile when _ENABLE_COMPILE is False
+    (CompilationMode.NONE, backend="eager", or config context unavailable).
+    """
+    if not _ENABLE_COMPILE:
+        return fn
+    return torch.compile(fn, dynamic=False)
```
tdoublep: Again, seems like a pretty common pattern. Couldn't this be imported some utils file instead of "mirroring the gating"?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -51,27 +58,228 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Compilable factory functions (PR #5 pattern)
```
tdoublep: What does "(PR #5 pattern)" mean?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -51,27 +58,228 @@
...
+            block_offset = block_offsets[t]
+            k_tok = convert(key_cpu[t].unsqueeze(1).contiguous(), target_device, target_dtype)
+            v_tok = convert(value_cpu[t].unsqueeze(1).contiguous(), target_device, target_dtype)
+            _overwrite(k_tok, k_pages[block_idx], [1], [block_offset])
+            _overwrite(v_tok, v_pages[block_idx], [1], [block_offset])
+
+    return specialized_reshape_and_cache_kernel
+
+
+def _create_compilable_page_attn(num_blocks: int, padded_query_len: int):
```
tdoublep, line 183: Why is query length padded but num_blocks not?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -117,37 +338,172 @@ def __init__(
...
         self.block_size = kv_cache_spec.block_size
+        self.head_size = kv_cache_spec.head_size
 
         model_config = vllm_config.model_config
         self.num_heads = model_config.get_num_attention_heads(vllm_config.parallel_config)
         self.num_kv_heads = model_config.get_num_kv_heads(vllm_config.parallel_config)
 
+        # Device is passed from model runner (spyre or cpu)
+        self._target_device = device
+        self._target_dtype = torch.float16
```
tdoublep: Do we really want this hard-coded?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -117,37 +338,172 @@ def __init__(
...
+            - mask_4d: [num_seqs * num_kv_heads, 1, padded_query_len, aligned_max_seq_len]
+            - padded_query_lens: [num_seqs] per-sequence padded query lengths
+        """
+        query_lens = query_start_loc[1:] - query_start_loc[:-1]
+        num_seqs = len(seq_lens)
+        num_kv_heads = self.num_kv_heads
+
+        # Compute per-sequence padded query length
+        padded_query_lens = (
+            (query_lens + QUERY_CHUNK_SIZE - 1) // QUERY_CHUNK_SIZE * QUERY_CHUNK_SIZE
```
tdoublep: Consider utils some utils function here? I think vllm has some built in

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -90,21 +298,34 @@ class SpyreAttentionMetadata(AttentionMetadata):
...
     num_heads: int = 0
 
+    # Pre-tiled attention mask: attention_mask_tiles[seq_idx][block_idx]
+    # Each tile: [num_kv_heads, 1, padded_query_len, block_size] on target device
+    attention_mask_tiles: list[list[torch.Tensor]] | None = None
+
+    # Per-sequence padded query length for stable kernel compilation
+    padded_query_lens: torch.Tensor | None = None  # [num_seqs] on CPU
+
+    aligned_max_seq_len: int = 0
```
tdoublep, line 321: Add comment to explain this variable? What is it aligned to?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -117,37 +338,172 @@ def __init__(
...
+        query_start_loc = common_attn_metadata.query_start_loc
+        max_seq_len = common_attn_metadata.max_seq_len
+        max_query_len = common_attn_metadata.max_query_len
+        block_table = common_attn_metadata.block_table_tensor
+        slot_mapping = common_attn_metadata.slot_mapping
+
+        apply_causal_mask = common_attn_metadata.causal and max_query_len > 1
+
+        aligned_max_seq_len = (
+            (max_seq_len + KV_LENGTH_ALIGNMENT - 1) // KV_LENGTH_ALIGNMENT * KV_LENGTH_ALIGNMENT
```
tdoublep, line 430: Again consider using pre-existing utils function here

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -117,37 +338,172 @@ def __init__(
...
+        )
+
+        mask_4d_cpu, padded_query_lens = self._build_attention_mask(
```
tdoublep: It looks like we don't really use the attention mask in this "4d" format. Would it make more sense to return it in the format that we actually want it.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -183,13 +538,16 @@ def get_kv_cache_shape(
         head_size: int,
         cache_dtype_str: str = "auto",
     ) -> tuple[int, ...]:
-        """KV cache shape: [num_blocks, 2, block_size, num_kv_heads, head_size]"""
-        return (num_blocks, 2, block_size, num_kv_heads, head_size)
+        # TODO this should be:
+        #     return (num_blocks, 2, block_size, num_kv_heads, head_size)
+        # but for the lists it is actually
```
tdoublep: Let's remove this TODO, I think it can be confusing for the reader.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -183,13 +538,16 @@ def get_kv_cache_shape(
...
+        return [  # ty: ignore[invalid-return-type]
+            (num_blocks, block_size, num_kv_heads, head_size),
+            (num_blocks, block_size, num_kv_heads, head_size),
+        ]
 
     @classmethod
     def supports_head_size(cls, head_size: int) -> bool:
-        # Spyre stick size is 128 bytes; tensors are transferred as float16 (2 bytes),
-        # so head_size must be a multiple of 64 (= 128 / 2) to satisfy stick alignment.
```
tdoublep, line 192: Is this no longer true, why remove?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -200,7 +558,16 @@ def supports_kv_cache_dtype(cls, kv_cache_dtype: CacheDType | None) -> bool:
...
 class SpyreAttentionImpl(AttentionImpl[SpyreAttentionMetadata]):
-    """PyTorch native implementation of attention with paged KV cache on Spyre."""
+    """Online-softmax paged attention iterating over KV pages.
+
+    KV cache is a tuple (k_pages, v_pages) where each is a list of tensors
+    of shape [num_kv_heads, block_size, head_size] on Spyre. No monolithic
+    cache tensor, no gather masks.
+
+    On Spyre, the per-page attention loop and reshape_and_cache are compiled
+    via torch.compile with fixed iteration counts (PR #5 pattern). A dict
```
tdoublep: see earlier comment re: (PR #5 pattern

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
         self.kv_cache_dtype = kv_cache_dtype
         self.attn_type = attn_type
 
-        # Target device/dtype for compiled attention kernels
-        self._target_device = torch.device("spyre")
+        self._target_device: torch.device | None = None
```
tdoublep: Why do we set to None?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
-        self.use_sdpa = use_sdpa
+        # Compiled function caches (keyed by iteration count)
+        self._reshape_fns: dict[int, object] = {}
+        self._attn_fns: dict[int, object] = {}
 
-        if self.use_sdpa:
-            self.attn_op = torch.nn.functional.scaled_dot_product_attention
-        else:
-            self.attn_op = _attn_4d
+        print("Using SpyreAttentionBackend with LIST-BASED online softmax (v0)")
```
tdoublep: Please use logger instead of prints

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
+        return self._attn_fns[key]
+
     def forward(
         self,
         layer: torch.nn.Module,
         query: torch.Tensor,  # [num_tokens, num_heads, head_size]
         key: torch.Tensor,  # [num_tokens, num_kv_heads, head_size]
         value: torch.Tensor,  # [num_tokens, num_kv_heads, head_size]
-        kv_cache: torch.Tensor,  # [num_blocks, 2, block_size, num_kv_heads, head_size]
+        kv_cache: tuple[list[torch.Tensor], list[torch.Tensor]],
```
tdoublep, line 609: Isn't it a list of tuples?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
-            attn_metadata.seq_lens,
-            attn_metadata.query_start_loc,
-            attn_metadata.apply_causal_mask,
-            attn_metadata.max_seq_len,
-            attn_metadata.max_query_len,
-            query.device,
+        assert (
+            attn_metadata.slot_block_indices is not None
+            and attn_metadata.slot_block_offsets is not None
+        ), "slot_block_indices/offsets must be precomputed by the metadata builder"
```
tdoublep: Do we need to assert everything to check that the metadata builder created it? Why these specifically?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
+        # bring q/k/v to CPU once for all slicing below; per-token slices get
+        # transferred to Spyre individually inside the scatter and attention paths.
+        key_cpu = convert(key, "cpu")
+        value_cpu = convert(value, "cpu")
+        query_cpu = convert(query, "cpu")
+
+        # Step 1: Reshape and cache — write new tokens into pages
+        self._reshape_and_cache(
+            key_cpu[:num_actual_tokens],
+            value_cpu[:num_actual_tokens],
```
tdoublep, line 636: Why do we slice by num_actual_tokens here? I thought the num_actual_tokens is something related to piecewise CUDA graphs which we are not using yet.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
+            q_start = int(query_start_loc[seq_idx].item())
+            q_end = int(query_start_loc[seq_idx + 1].item())
+            query_len = q_end - q_start
+            kv_len = int(seq_lens[seq_idx].item())
+
+            # Use per-sequence padded query length
+            padded_query_len = int(padded_query_lens[seq_idx].item())
+
+            query_cpu = convert(query, device="cpu")
+            q_seq = convert(query_cpu[q_start:q_end], device=self._target_device)
```
tdoublep: It seems like we have quite some churn moving things back from CPU to Spyre and back again, Why not pass query in on CPU?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +590,212 @@ def __init__(
...
+            # handling the degenerate dimension in MHA. GQA (num_queries_per_kv > 1)
+            # works correctly. See error: "cannot restickify any input layout of y
+            # to carry y_var=d2" in propagate_layouts.py:341
+
+            num_blocks_needed = (kv_len + block_size - 1) // block_size
+            page_indices = [int(block_table[seq_idx, i]) for i in range(num_blocks_needed)]
+            mask_tiles = [m.to(self._target_device) for m in mask_tiles_all[seq_idx]]
+
+            # Run attention on target device
+            q_dev = convert(q, self._target_device, self._target_dtype)
```
tdoublep: Isnt' this already on device?

spyre_inference/v1/worker/spyre_worker.py
```diff
@@ -96,7 +96,7 @@ def compile_or_warm_up_model(self) -> CompilationTimes:
         import time
 
         import torch._inductor.decomposition
-        from torch_spyre._inductor.decompositions import spyre_decompositions  # ty: ignore[unresolved-import]
+        from torch_spyre._inductor.decompositions import spyre_decompositions
```
tdoublep, line 99: Why this change is needed?

tests/conftest.py
tdoublep, line 1: Why are these changes needed?

tests/test_spyre_attn.py
```diff
@@ -295,34 +370,39 @@ def test_spyre_attn(
...
         scale=scale,
         sliding_window=sliding_window,
         soft_cap=soft_cap,
     )
 
-    if max_query_len >= 32:
-        atol, rtol = 0.3, 5.0
+    if max(query_lens) >= 32:
+        # TODO: what values are sensible here?
+        atol, rtol = 0.9, 5.0
```
tdoublep: Now we have fixed bugs can we use lower tolerance here?

tdoublep (commented):
Great work!

examples/offline_inference/torch_spyre_inference.py
bringlein: are you sure? I could not get it to work without it

examples/offline_inference/torch_spyre_inference.py
bringlein: yes, but you need to select it then?

examples/offline_inference/torch_spyre_inference.py
bohnstingl: No, I don't think this option makes any sense anymore. It is outdated, I removed it.

examples/offline_inference/torch_spyre_inference.py
bohnstingl: Removed.

examples/offline_inference/torch_spyre_inference.py
bohnstingl: There is no point of this anymore. It is also a relict from the past. Actually @R3hankhan123 is working on enforcing the CUSTOM attention backend in the platform, see https://github.com/torch-spyre/spyre-inference/pull/231. We can then remove the argument from here.

examples/offline_inference/torch_spyre_inference.py
bohnstingl: Removed.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I refactored the code to mirror the behavior used in CustomOps in upstream vLLM.

Just as a side note, one important consideration here is that we enforce `opaque_attention_op` to return `False` in our plugin. This was necessary in order to keep the attention output on spyre, see.
However, this also means that when using `enforce_eager=False`, the attention function will be tried to be compiled, which causes issues with our D2H and H2D data transfers. To ensure that everything is working properly with enforce_eager=True and False, I put a

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I consolidated, see comment above.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I think this is not necessary. We can pick the device from the attention inputs. I changed that. We don't change the `dtype`, so this is not necessary in the first place. In the `_build_attention_mask` function, we now pick the `dtype` from the model itself.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Removed.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Changed.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: No, it is currently a tuple of lists, see its allocation [here](https://github.com/bringlein/spyre-inference/blob/dad21091b895cc4e520c889b9e7828e903076540/spyre_inference/v1/worker/spyre_model_runner.py#L349-L370). We can change it though in any way we want.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I don't think that this is a necessity, but @bringlein maybe you had a good reason for it?

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I am not sure if the `num_actual_tokens` is strictly tied to CUDA graphs. The particular slicing appears to be a common pattern in several places, including the [CPU attention](https://github.com/vllm-project/vllm/blob/68f5e565c90b670911024a2558aa265575b3e71a/vllm/v1/attention/backends/cpu_attn.py#L332). However, it is also used in other places such as the [Triton attention](https://github.com/vllm-project/vllm/blob/68f5e565c90b670911024a2558aa265575b3e71a/vllm/v1/attention/backends/triton_attn.py#L635), or the [FlashAttention](https://github.com/vllm-project/vllm/blob/68f5e565c90b670911024a2558aa265575b3e71a/vllm/v1/attention/backends/flash_attn.py#L725).

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Yeah, initially we thought that we keep as many operations on spyre as possible. However, in this specific case it really creates unnecessary overhead. I reverted and pass the query in on CPU.

spyre_inference/v1/worker/spyre_worker.py
bohnstingl: Running `bash format.sh` places this

bohnstingl:
I haven't finished with all my changes yet. But I wanted to enable @bringlein and thus pushed my current changes which should address several of @tdoublep requests. Not yet done fighting with the CI tests.

FYI, I am deferring the enablement of the `torch.compile` mode for the attention backend to a follow-up PR, in order not to block this one further.

examples/offline_inference/torch_spyre_inference.py
bohnstingl: On my side it is working without this.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Removed.

examples/offline_inference/torch_spyre_inference.py
bringlein: no, for me I had to add `VLLM_PLUGINS=spyre_inference` again, otherwise it runs on CPU. I guess this is because my pod doesn't have it in the environment. Hence, I think it is better to do it here to be safe.

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: I guess not necessarily...

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: the query length was padded before and yes `num_blocks` should be padded in a future PR, I guess we should not include it in this one.

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: it referred to the list-based pattern explored in PR #5 ....but yes, maybe not necessary as comment.

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: I haven't found one...because we would need to align it with the query chunk size? Can you point me to what you have in mind?

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: added comments

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: same comment as above, can you point me to one?

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: that's right, I removed the unnecessary masks

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: No, I also don't think so :D, cleaned it up

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: but with the pre-compiled function we would end up using something very similar?

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: due to slicing sadly not? See discussion above

tests/conftest.py
bringlein: I couldn't get the unit tests to work without, but apparently this was my stupidity...

bringlein: ok, update: I forgot to install/re-install the tests plugin and that's why the fixtures for the pytest were not discovered (and hence the agent added the conftest to point to it). 
So, 
```
SETUPTOOLS_SCM_PRETEND_VERSION=0.0.1 uv pip install -e tests/plugin/ --no-deps
```
fixes this and `conftest.py` is not necessary.

spyre_inference/v1/attention/backends/spyre_attn.py
tdoublep:
```
from vllm.utils.math_utils import round_up
```

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -224,446 +572,201 @@ def __init__(
...
-            self.attn_op = torch.nn.functional.scaled_dot_product_attention
-        else:
-            self.attn_op = _attn_4d
+        # Compiled function caches (keyed by iteration count)
+        self._reshape_fns: dict[int, object] = {}
+        self._attn_fns: dict[int, object] = {}
 
-        # Compile the attention function once for reuse.
-        self.attn_op = torch.compile(self.attn_op)
+        logger.debug_once("Using SpyreAttentionBackend with LIST-BASED online softmax (v0)")
```
tdoublep: Let's not put arbitrary version number in logs like that

spyre_inference/v1/attention/backends/spyre_attn.py
tdoublep: My bad. I was confusing it with what we return from `get_kv_cache_shape` above

spyre_inference/v1/attention/backends/spyre_attn.py
tdoublep: Does the slicing actually have any effect at this point in time?

examples/offline_inference/torch_spyre_inference.py
tdoublep: The only thing necessary should be:
```
VLLM_PLUGINS=spyre_inference
```
It should be in the dev pod, but I'm OK with also explicitly setting it in the example. The `VLLM_TARGET_DEVICE` does definitely not need to be set though, neither the `spyre_inference_ops`.

bohnstingl:
I've pushed a commit that increases the `block_size` in the attention tests to make them pass. There are issues on the torch-spyre side with smaller block_sizes, I have created and issue tracking this, see https://github.com/torch-spyre/spyre-inference/issues/239.

tests/test_spyre_attn.py
```diff
@@ -244,7 +244,7 @@ def ref_attn(
 @pytest.mark.parametrize(
     "block_size",
     [
-        pytest.param(16, id="block_size(16)"),
+        pytest.param(128, id="block_size(128)"),
```
bringlein, line 315: ok, I feel again quite stupid to have overlooked this 🙈 🤦 😆

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: fixed by BOH

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -622,6 +622,8 @@ def forward(
         # (e.g. in unit tests) while pages live on the real Spyre device.
         _target_device = k_pages[0].device
         num_actual_tokens = attn_metadata.num_actual_tokens
+        assert attn_metadata.slot_block_indices is not None
+        assert attn_metadata.slot_block_offsets is not None
```
bringlein: @bohnstingl why do we need this? from linting?

examples/offline_inference/torch_spyre_inference.py
bohnstingl: As discussed offline, this might be a potential pod issue?

tests/test_spyre_attn.py
bringlein: on the contrary, it did work above...before the changes necessary to run end2end?

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Yeah, they have been introduced by the linting. I found another way though and removed them again

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: no, since we don't pad, I would say not. But since we want to pad very soon...I would keep it? Or should we remove it and save one CPU-Spyre cycle?

examples/offline_inference/torch_spyre_inference.py
bringlein: > It should be in the dev pod, but I'm OK with also explicitly setting it in the example. The VLLM_TARGET_DEVICE does definitely not need to be set though, neither the spyre_inference_ops.

yes, I agree and can confirm, `VLLM_PLUGINS=spyre_inference` is enough.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Removed the v0.

tests/test_spyre_attn.py
bohnstingl: No worries, for the moment this is the solution. I've raised a follow-up issue though, see https://github.com/torch-spyre/spyre-inference/issues/239

tests/test_spyre_attn.py
bringlein: yes, we can lower atol to 0.55 here in case of prefill, but below would result in errors like:
```
Mismatched elements: 3 / 131072
```

Hence I created a quick helper that would allow e.g. max 5 "outliers" with twice the atol. And with that we can set the atol to 0.3 for prefill (0.2 for decode). I guess this is the best solution / most-accurate test?

bringlein:
I lowered the tolerance in the `assert_close` again and did some cleaning together with @bohnstingl . 
It looks like the CI infra has a problem, but for me locally the tests pass

```
SKIP_UPSTREAM_TESTS=1 pytest tests/test_spyre_attn.py -x -rsx -v
....
24 passed, 8 skipped, 29 warnings in 262.62s (0:04:22)
```

And also end-to-end, I get correct output:
```
python3 examples/offline_inference/torch_spyre_inference.py --model ../../models/ibm-granite/granite-3.3-8b-instruct/ --max-model-len 1024 --max-num-seqs 1 --num-gpu-blocks-override 64 --num-prompts 1

.... 
                                                                                                                                                                                         
CompletionOutput(index=0, text="\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,", token_ids=[203, 203, 36671, 1182, 2575, 6866, 46654, 2305, 6564, 219
09, 11258, 30, 6564, 17265, 11258, 30, 3859, 30, 18919, 30], routed_experts=None, cumulative_logprob=None, logprobs=None, finish_reason=length, stop_reason=None)                                                
===============                                                                                                                                                                                                  
                                                                                                                                                                                                                 
Prompt:                                                                                                                                                                                                          
 'What are IBMs main businesses?'                                                                                                                                                                                
                                                                                                                                                                                                                 
Generated text:                                                                                                                                                                                                  
 "\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,"  
```

So I guess we wait for CI infra and then are good to go?

tdoublep (approved):
LGTM

bohnstingl:
bot:next-test

bohnstingl (approved):

bohnstingl:
@joerunde I think the Spyre CI tests actually pass? Do we understand why the "test_each_commit" tests fail?

PR #209, opened 2026-05-22
:sparkles: Add TP=2 support for vocab embeddings

### LINKED ISSUE #135: Add tp>1 support for Embed Layer

joerunde:
## Description

This PR brings back the `SpyreVocabParallelEmbeddings` custom op for TP support. This is required because there are int64 comparison checks in [get_masked_input_and_mask](https://github.com/vllm-project/vllm/blob/v0.21.0/vllm/model_executor/layers/vocab_parallel_embedding.py#L163-L187) when building the mask over the vocabulary for each shard in a TP configuration. Spyre does not support integer math at all right now, so we must pull the input tensor back to the cpu to do the masking, then send the masked tensor back to the device.

Once spyre _does_ support integer math, we can probably remove this custom op again. To that end, I've added a test with a strict xfail in `test_int64_compiled_compare_against_python_int` that checks for integer math support on spyre devices. Once that's supported the test will fail, alerting us that we can come back and remove this op.

## Related Issues

Closes #135 

## Test Plan

See unit tests

## Checklist


joerunde:
bot:next-test
🤞 on the new base image for testing

joerunde:
bot:next-test

Just checking latest main merge

tests/test_distributed_tp2.py
```diff
@@ -59,6 +59,28 @@ def test_tp2_tensor_model_parallel_all_reduce(run_tp_probe) -> None:
...
+    real init path, constructs the OOT VocabParallelEmbedding, and
+    asserts each rank's all-reduced output matches the full-vocab
+    F.embedding reference. Independent of #134 (linear TP).
+
+    Isolates the embedding-TP correctness check from #134 (linear TP) and
+    #136-equivalent (LM head TP) so the embedding work can land
+    independently. When the full LLM(tp=2) test passes, this is
+    redundant and can be deleted.
+    """
+    run_tp_probe("vocab_parallel_embedding", world_size=2)
```
prashantgupta24, line 81: could this instead be combined with `test_tp2_tensor_model_parallel_all_reduce`, something like

```
  @pytest.mark.parametrize("probe_name", [
      "tp_all_reduce",
      "vocab_parallel_embedding",
  ])
def test_tp2_distributed_ops(run_tp_probe, probe_name: str) -> None:
      ...
      run_tp_probe(probe_name, world_size=2)
```

prashantgupta24: And we add probes later as they are added

joerunde: yeah we could, I just split these out so that they could each have their own relevant docstring. I can try inlining the important bits with the parametrization though

joerunde: ah- actually these are both probably about to be deleted once @R3hankhan123's queued up PRs land so to avoid churn i'd like to just leave as-is

prashantgupta24 (approved):

maxdebayser (approved):

PR #213, opened 2026-05-27
Edits of existing attention backends

bohnstingl:
## Description

This PR makes several modifications and adds some new features to the existing attention backends. In particular, it

- Ensures that the attention `output` is on spyre
- Ensures that as many tensors as possible arrive on spyre for the attention (KV cache, key tensor, value tensor, output)
- It allocates the KV caches for both backends in the SpyreModelRunner (single KV cache tensor for `spyre_attn.py` and tuple for `spyre_attn_exp.py`)
- It reworks `torch.ops.spyre.overwrite_f` with `torch.ops.spyre.overwrite`
- It consolidates the `_attn_4d` function, the `_maybe_compile` function, the `output` copying of the two backends into a utils function

## Related Issues

This PR is related to https://github.com/torch-spyre/spyre-inference/issues/121

## Test Plan

The changes of this PR are transparent to the tests, so the existing tests can be reused.

## Checklist


spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -320,10 +321,65 @@ def warming_up_model(self) -> None:
...
+            layer_names = set()
+            for group in kv_cache_config.kv_cache_groups:
+                for layer_name in group.layer_names:
+                    if layer_name in self.runner_only_attn_layers:
+                        continue
+                    layer_names.add(layer_name)
+            assert layer_names == set(kv_cache_raw_tensors.keys()), (
+                "Some layers are not correctly initialized"
+            )
+            return kv_cache_raw_tensors
```
bohnstingl: This is similar code from upstream `vLLM` and creates the two KV cache tensors that the _exp attention backend requires.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -268,6 +254,12 @@ def forward(
         if attn_metadata is None:
             return output
 
+        # The value still resides on CPU, as it is
+        # created in the GraniteAttention through a split
```
tdoublep, line 271: Is this limitation really specific to a specific model (e.g., Granite)? 

The comment also seems to be about the value tensor, whereas the code below relates to the query and key. Is the implication that we need to move the query and key to the CPU because the value resides there? Why not move the value to Spyre?

spyre_inference/v1/attention/backends/spyre_attn_exp.py
```diff
@@ -641,24 +594,21 @@ def _scatter_to_device_cache(
             # layout doesn't match what overwrite_f's bundle compiler
             # expects (triggers patch (A)'s narrowed-view guard). A
             # fresh CPU->Spyre upload reshapes cleanly.
-            k_cpu = key.to(self._target_dtype).contiguous()
-            v_cpu = value.to(self._target_dtype).contiguous()
-
-            k_cache = self._k_cache_dev
-            v_cache = self._v_cache_dev
+            k_cpu = convert(key, dtype=self._target_dtype)
+            v_cpu = convert(value, dtype=self._target_dtype)
```
tdoublep: What the fundamental reason we need to do this processing on the CPU?

spyre_inference/v1/worker/spyre_model_runner.py
jvlunteren: The return type says `dict[str, torch.Tensor]` but for the exp path you're storing tuples (k_cache_dev, v_cache_dev). Is that correct?

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: Well, to be accurate this is an issue with the upstream vLLM code of Granite, but potentially also of other models. [GraniteAttention](https://github.com/vllm-project/vllm/blob/52a31ccecca203effd490a5b496dc5f8d9496654/vllm/model_executor/models/granite.py#L106), performs a split operation, see https://github.com/vllm-project/vllm/blob/52a31ccecca203effd490a5b496dc5f8d9496654/vllm/model_executor/models/granite.py#L182. This split operation has currently to be executed on `cpu`. In the `spyre` implementation of RoPE we put q and k back to `spyre`, but v is not touched before the attention implementation anymore, thus it remains on `cpu`.
So in conclusion, `k` and `q` arrive to the attention backend on spyre, while the `v` arrives on cpu.

Inside the attention implementation we need to put `k` and `q` onto cpu because we perform tensor slicing, which today only works on cpu.

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -320,10 +321,61 @@ def warming_up_model(self) -> None:
...
+        self,
+        kv_cache_config,
+    ) -> dict[str, torch.Tensor | tuple[torch.Tensor, torch.Tensor]]:
+        if envs.SPYRE_ATTN_IMPL == "exp":
+            kv_cache_raw_tensors: dict[str, tuple[torch.Tensor, torch.Tensor]] = {}
+            for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
+                block_size = self.vllm_config.cache_config.block_size
+                num_blocks = kv_cache_config.num_blocks
+                head_size = self.vllm_config.model_config.model_arch_config.head_size
+                num_kv_heads = self.vllm_config.model_config.model_arch_config.total_num_kv_heads
```
bohnstingl, line 334: This might be a somewhat abusive way of collecting the values to prepare the cache tensors. However, since the `_exp` backend might get deprecated at some point, we could just leave it for the moment?

In general, the code paths with ```if envs.SPYRE_ATTN_IMPL == "exp"``` will be removed.

spyre_inference/v1/worker/spyre_model_runner.py
bohnstingl: Good point, I fixed the type annotation

spyre_inference/v1/attention/backends/spyre_attn_exp.py
bohnstingl: These two particular lines are not required and I removed them. The `key` and the `value` tensor are anyways already on `cpu`. The reason we have to keep them on cpu is that in the lines below, see [here](https://github.com/bohnstingl/spyre-inference/blob/f65427ad85264693ffaa5b6c44a476cd735bfa84/spyre_inference/v1/attention/backends/spyre_attn_exp.py#L608-L609), we do a slicing operation.

bohnstingl:
bot:next-test

spyre_inference/v1/attention/backends/spyre_attn_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    cfg = get_cached_compilation_config()
+    if cfg.mode == CompilationMode.NONE:
+        return fn
+    if cfg.backend == "eager":
+        return fn
+    return torch.compile(fn)
+
+
+def _attn_4d(q, k, v, scale, mask):
```
bringlein: Is there a reason we move this hear? I guess in the medium/long run we remove it?

spyre_inference/v1/attention/backends/spyre_attn_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    """Place `attn_output[:num_actual_tokens]` into the caller-provided `output` buffer."""
+    if output.device.type == "spyre":
+        # Workaround for torch-spyre bug: a CPU->Spyre `.copy_()` into a
+        # destination whose rank differs from its allocation rank (vLLM
+        # allocates `output` 2-D and views it 3-D) crashes in
+        # spyre::get_device_stride_infos. Build the full result on CPU,
+        # upload as a fresh Spyre tensor, then Spyre->Spyre copy (which
+        # dispatches through torch.ops.spyre.copy_from_d2d and works).
+        full_cpu = torch.zeros(output.shape, dtype=output.dtype, device="cpu")
+        full_cpu[:num_actual_tokens] = attn_output
```
bringlein: In the draft PR #206 I try to use `torch.ops.spyre.overwrite` for this too...wouldn't that save one CPU-Spyre cycle? 
https://github.com/bringlein/spyre-inference/blob/9b3592b762df143e83814b1d4221765414e9e3cf/spyre_inference/v1/attention/backends/spyre_attn.py#L628-L634

spyre_inference/v1/worker/spyre_model_runner.py
bringlein: You mean to have separate tensors for k and v? I think that might be a requirement we have to keep for the absolute address calculation.

bohnstingl: Yes, the separate `k_cache_dev` and `v_cache_dev` were already present in initial merge https://github.com/torch-spyre/spyre-inference/pull/132. In particular, this PR moved the cache tensors from [here](https://github.com/jvlunteren/spyre-inference/blob/9b4fe9544a4b1cfc7fe80ba3733f326cf84064db/spyre_inference/v1/attention/backends/spyre_attn_exp.py#L510-L523), into the ModelRunner.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bohnstingl: I consolidated the `_attn_4d` function into a utils file, because the same function is used in [spyre_attn.py](https://github.com/bohnstingl/spyre-inference/blob/35573c957254795d846241467b051f737cfdd7f8/spyre_inference/v1/attention/backends/spyre_attn.py#L225) and in [spyre_attn_exp.py](https://github.com/bohnstingl/spyre-inference/blob/35573c957254795d846241467b051f737cfdd7f8/spyre_inference/v1/attention/backends/spyre_attn_exp.py#L463).

I agree that in the long-run we may only have a single backend and no need for the utils.py. If you rather prefer not to have the utils function at all, that's fine with me as well.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bohnstingl: Yes, this is a similar usecase. The current code, for example for the k-cache

```
k_tok = convert(key[t].unsqueeze(1).contiguous(), device=self._target_device)
torch.ops.spyre.overwrite(k_tok, self._k_cache_dev, [1, 2], [br, cs])
```

converts the `key[t]` once to `spyre` and then performs the operation on `spyre`. At least this is my understanding. The original `key` needs to be on cpu, because of the slicing operation.
What would be your suggestion to improve?

tdoublep (approved):

spyre_inference/v1/attention/backends/spyre_attn_utils.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    from vllm.config import get_cached_compilation_config
+    from vllm.config.compilation import CompilationMode
+
+    cfg = get_cached_compilation_config()
+    if cfg.mode == CompilationMode.NONE:
+        return fn
+    if cfg.backend == "eager":
+        return fn
+    return torch.compile(fn)
```
joerunde: should `dynamic=False` be in here? It was on the previous torch.compile call sites

spyre_inference/v1/attention/backends/spyre_attn_utils.py
joerunde: All three of these methods are used only by our `AttentionImpl`s, so inheritance seems like the more natural pattern to use here instead of a utils file.

Here's a sketch of how this could be refactored, also converting `maybe_compile` to a decorator to decorate an inherited `attn_4d` static method. This would let us cleanly delete `spyre_attn_exp.py` later

```python

 # spyre_attn.py

  import functools
  import torch

  
  def _maybe_compile(fn):
      """Lazily wrap `fn` with torch.compile on first call.

      Compile gating is checked at first call, not at decoration time, so
      the runtime CompilationConfig wins over whatever was in effect at
      class import.
      """
      compiled = None

      @functools.wraps(fn)
      def wrapper(*args, **kwargs):
          nonlocal compiled
          if compiled is None:
              from vllm.config import get_cached_compilation_config
              from vllm.config.compilation import CompilationMode

              cfg = get_cached_compilation_config()
              if cfg.mode == CompilationMode.NONE or cfg.backend == "eager":
                  compiled = fn
              else:
                  # dynamic=False: required by the Spyre compiler.
                  compiled = torch.compile(fn, dynamic=False)
          return compiled(*args, **kwargs)

      return wrapper


  class SpyreAttentionImpl(AttentionImpl[SpyreAttentionMetadata]):
  """PyTorch native implementation of attention with paged KV cache on Spyre."""

      @_maybe_compile
      @staticmethod
      def _attn_4d(q, k, v, scale, mask):
          # (same impl as today)

      @staticmethod
      def copy_attn_output(attn_output, num_actual_tokens, output):
         # (same impl as today)

    ...

# spyre_attn_exp.py

  from spyre_attn import SpyreAttnImpl

  class SpyreExpAttnImpl(SpyreAttnImpl):
    ...
```

bringlein: since we will have only one of the `spyre_attn` backends in the very near future, I think it is not worth trying to optimize this reuse here, IMHO.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bringlein: which tensor do you refer here? I meant only the attention output...to avoid additional spyre-cpu copies. But maybe I'm mistaken.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bohnstingl: No, this is not needed anymore. torch-spyre does support `dynamic=True` now as well.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bohnstingl: I was somehow wrongly pointed to `k_tok`, but I realized that your comment was about the `attn_output`, I introduced `torch.ops.spyre.overwrite` for this one as well.

spyre_inference/v1/attention/backends/spyre_attn_utils.py
bohnstingl: I can see both points. Thank you @joerunde for the tips for a better software development approach to this and thank you @bringlein for your opinion. I was going back and forth on this and actually since the target is to really have only one backend and remove the exp one very soon, on a second thought I also tend to not waste time / effort in any unification. I will keep the methods duplicated for the moment and not do any reuse optimization.

bohnstingl:
bot:next-test

bohnstingl:
@joerunde can you maybe help me check the CI results? From the console output it seems that the tests actually didn't even run?

joerunde:
bot:next-test

checking if the CI pods are even standing up

bohnstingl:
bot:next-test

bohnstingl:
@jvlunteren, @bringlein could you maybe take a look at this and see whether anything else strikes you odd?

bringlein (approved):
LGTM

bohnstingl:
bot:next-test

dilipgb (approved):

PR #216, opened 2026-06-02
[Custom Ops] Implement TP in linear layers

### LINKED ISSUE #134: Add tp>1 support for Linear Layers

R3hankhan123:
## Description
This PR adds support for tensor Parallelism in linear layers. Note once `all_gather` is supported `SpyreMergedColumnParallelLinear` can be removed as its a mirror copy of upstream vllm just adding the runtime warning of all_gather not supported.

all_gather not being supported is not a blocker for most LLMs as gather_output param in model code in vllm have set it to false. This will only effect vision models and MTP models like gemma4 and some qwen 3.5 models
## Related Issues
resolves #134 
## Test Plan
Unit tests have been written and are passing
## Checklist

### DIFF
--- a/spyre_inference/custom_ops/linear.py
+++ b/spyre_inference/custom_ops/linear.py
@@ -33,7 +33,7 @@ Spyre Device Constraints:
     - Computations performed in torch.float16:
       Input (dtype defined by model / user) converted to torch.float16 for
       operations on spyre and then converted back to original dtype for cpu.
-    - Tensor parallelism: TP=1 assumed (single Spyre device)
+    - Tensor parallelism: TP>=1 supported with all_reduce collectives
 
 References:
     - Upstream linear layers:   vllm/model_executor/layers/linear.py
@@ -42,6 +42,10 @@ References:
 import torch.nn.functional as F
 
 from vllm.logger import init_logger
+
+from spyre_inference.distributed.spyre_communicator import (
+    _spyre_collective_unsupported_message,
+)
 from vllm.model_executor.layers.linear import (
     MergedColumnParallelLinear,
     QKVParallelLinear,
@@ -58,7 +62,7 @@ class SpyreUnquantizedLinearMethod(UnquantizedLinearMethod):
     """Spyre-specific linear method: F.linear without platform GEMM dispatch.
 
     Replaces the default UnquantizedLinearMethod so that upstream forward()
-    methods work unchanged on Spyre at TP=1.
+    methods work unchanged on Spyre at any TP size.
 
     - create_weights() is inherited — standard ModelWeightParameter works.
     - apply() does F.linear directly (no platform-specific GEMM dispatch).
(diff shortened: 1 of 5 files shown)


spyre_inference/distributed/spyre_communicator.py
```diff
@@ -156,21 +156,6 @@ def all_reduce(self, input_: torch.Tensor) -> torch.Tensor:
         dist.broadcast(input_, src=self.ranks[0], group=self.device_group)
         return input_
 
-    def all_gather(self, input_: torch.Tensor, dim: int = -1) -> torch.Tensor:
```
R3hankhan123, line 159: This is required since its causing error on lm_head logits gathering

spyre_inference/custom_ops/linear.py
```diff
@@ -73,44 +82,130 @@ def process_weights_after_loading(self, layer):
...
+        """
+        bias = self.bias if not self.skip_bias_add else None
+
+        output_parallel = self.quant_method.apply(self, input_, bias)
+
+        if self.gather_output and self.tp_size > 1:
+            raise NotImplementedError(
+                _spyre_collective_unsupported_message(
+                    "allgather",
+                    self.world_size,
```
joerunde: self.tp_size here? Not sure self.world_size exists

spyre_inference/custom_ops/linear.py
```diff
@@ -73,44 +82,130 @@ def process_weights_after_loading(self, layer):
...
```
joerunde: self.tp_size?

spyre_inference/distributed/spyre_communicator.py
joerunde: Can we get some more info? all_gather isn't implemented yet by spyreccl, and removing this guard here would cause callers to hit a failure further down the stack

spyre_inference/custom_ops/linear.py
```diff
@@ -73,44 +82,130 @@ def process_weights_after_loading(self, layer):
...
         super().__init__(*args, **kwargs)
-        if self.tp_size > 1:
-            raise NotImplementedError(
-                f"{self.__class__.__name__} only supports TP=1, got TP={self.tp_size}"
-            )
 
         if isinstance(self.quant_method, UnquantizedLinearMethod):
             self.quant_method = SpyreUnquantizedLinearMethod()
 
+        logger.debug(
```
joerunde: this would probably be too chatty- isn't this gonna log for every linear layer in the model?

spyre_inference/custom_ops/linear.py
```diff
@@ -73,44 +82,130 @@ def process_weights_after_loading(self, layer):
...
+                    self.world_size,
+                    blocker="libspyre_comms list-form allgather + torch-spyre",
+                )
+            )
+
         # D2H before downstream .split() — Spyre can't handle strided views
-        if self.return_bias:
-            return convert(result[0], device="cpu"), result[1]
-        return convert(result, device="cpu")
+        output = convert(output_parallel, device="cpu")
```
joerunde: See suggestions here: https://github.com/R3hankhan123/spyre-inference/pull/6

I think we should have a test for when this extra transfer can be removed, and I think we can also use the super() calls for most of these - at least the current tests don't fail so we need more test coverage if there are other required workarounds in these .forward methods

R3hankhan123: Thanks @joerunde for the suggestions, I have incorporated the PR

spyre_inference/distributed/spyre_communicator.py
R3hankhan123: https://github.com/vllm-project/vllm/blob/f0204358d9b811bde5320c037236e97b8fb6199d/vllm/platforms/interface.py#L851
ok looks like this is setting it to true

R3hankhan123: And using this value it determines weather to use all gather or gather
https://github.com/vllm-project/vllm/blob/f0204358d9b811bde5320c037236e97b8fb6199d/vllm/model_executor/layers/logits_processor.py#L75-L87

R3hankhan123: looks like these logits are on CPU thats why its not failing 
```
(Worker_TP0 pid=121113) INFO 06-03 06:26:19 [logits_processor.py:81] [DEBUG _gather_logits] logits device: cpu, dtype: torch.float16, shape: torch.Size([1, 24608])
(Worker_TP0 pid=121113) INFO 06-03 06:26:19 [logits_processor.py:85] [DEBUG _gather_logits] use_all_gather: True
(Worker_TP0 pid=121113) INFO 06-03 06:26:19 [logits_processor.py:93] [DEBUG _gather_logits] Calling tensor_model_parallel_all_gather()
(Worker_TP0 pid=121113) INFO 06-03 06:26:19 [logits_processor.py:101] [DEBUG _gather_logits] Result device: cpu, shape: torch.Size([1, 49216])
```

joerunde: Interesting!

It looks like our previous guard here was just too broad- we can check the dtype of the tensors and if they're on cpu we can allow the super() impl to take care of the gather / all_gather. I've reinstated this with that change and it simplifies linear.py a lot too

joerunde (approved):
(speculative ✅ on changes that are half mine anyway 😉 )

spyre_inference/custom_ops/linear.py
```diff
@@ -73,43 +74,69 @@ def process_weights_after_loading(self, layer):
...
+            "all_gather is not yet supported on Spyre"
+        )
 
     def forward(self, input_):
         result = super().forward(input_)
-        # D2H before downstream .split() — Spyre can't handle strided views
+        # D2H so that GraniteAttention's qkv.split() and the subsequent
+        # v.view() + kv_cache scatter-write run on CPU. Spyre rejects a
+        # non-contiguous tensor as a scatter source; see
+        # test_spyre_strided_scatter_source for the minimal reproduction.
```
bohnstingl, line 125: We may not need to expand this really?

spyre_inference/custom_ops/linear.py
```diff
@@ -73,43 +74,69 @@ def process_weights_after_loading(self, layer):
...
-    `input_` is on `cpu` and from `GraniteMLP` where `input_` is on spyre.
-    Thus, we always convert the `input_` to `spyre`, which is a NoOp in
-    case of `GraniteMLP`.
+    """Spyre RowParallelLinear with TP support.
+
+    Supports TP>=1 with weight sharding along input dimension and
+    all_reduce for aggregating results across ranks when reduce_results=True.
+
+    RowParallelLinear is invoked from GraniteAttention (input on cpu) and
+    GraniteMLP (input on spyre); H2D is a no-op in the latter case.
```
bohnstingl, line 139: I think this comment may not be needed?

spyre_inference/distributed/spyre_communicator.py
```diff
@@ -157,12 +157,17 @@ def all_reduce(self, input_: torch.Tensor) -> torch.Tensor:
...
         # SpyreCommsContext::allgather (and torch-spyre's spyreccl backend
         # wires up `_allgather_base`), delete this override and let the
         # base class use `dist.all_gather_into_tensor`.
         if self.world_size == 1:
             return input_
+        if input_.device.type == "cpu":
+            return super().all_gather(input_, dim)
         raise NotImplementedError(
             _spyre_collective_unsupported_message(
                 "allgather",
```
bohnstingl, line 173: I am just a little confused by this here. Doesn't tp > 1 on spyre involve `all_gather` and `gather`? From this snippet it seems that it is actually not supported on spyre?

bohnstingl (changes requested):

spyre_inference/distributed/spyre_communicator.py
R3hankhan123: Yes for now `all_gather` and `gather` is not supported for spyre but since logits are on CPU it doesn't fail and `tensor_model_parallel_all_gather` falls back to cpu

bohnstingl: Okay, so this PR prepares the ground for TP>1 once, we have all_gather and gather really supported on spyre, right?

spyre_inference/custom_ops/linear.py
joerunde: might be too much info, sure. The idea here was though that we could remove this cpu transfer once that test starts passing

spyre_inference/distributed/spyre_communicator.py
joerunde: pretty much!

With the next PR to address lm_head we should at least be able to run TP=2 with the fallback send/recv and the cpu transfer here on SpyreQKVParallelLinear

spyre_inference/custom_ops/linear.py
joerunde: yeah we could remove this entire op probably- just left it here for more context for the reader and future AI agents. I don't have a strong feeling either way

bohnstingl (approved):
Given the known limitations, fine with me

PR #220, opened 2026-06-03
cicd: add vLLM benchmark CI for spyre

YashasviChaurasia:
## Description

Minimal initial workflow: latency + throughput tests for validation of workflow
currently uses dummy weights for fast validation on spyre_pf_x1 runners. 
Includes ClickHouse ingestion workflow triggered on completion.

## Checklist


joerunde (changes requested):
See related review comments on https://github.com/torch-spyre/spyre-inference/pull/196

Using pre-baked environments or installs adds way too much complexity here, breaks isolation between builds, and causes us to build with different sets of dependencies than the checkout actually specifies.

There shouldn't be any good reason that the full checkout and build process can't just be
```
    steps:
    - name: "Checkout"
      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
    - name: "Install uv"
      uses: astral-sh/setup-uv@v5
    - name: "Install package"
       run: uv sync --frozen
```

with the appropriate environment variables set up for `UV_CACHE_DIR` and `HF_HOME`/`HF_HUB_CACHE` so that we cache to a shared pvc

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
```diff
@@ -0,0 +1,150 @@
...
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            .github/scripts/ingest_vllm_benchmarks.py
+            .github/scripts/requirements-benchmark.txt
+
+      - name: Install Python dependencies
+        run: |
+          source /home/senuser/spyre-inference/.venv/bin/activate 2>/dev/null || true
+          pip install clickhouse-connect requests
```
joerunde: Did this intend to use `requirements-benchmark.txt` (which seems unused?)

Also GitPython is listed on there, but not here

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,212 @@
...
+
+          # Run latency benchmarks directly (Spyre — no GPU wrapper needed)
+          if [ -f "$TESTS_DIR/latency-tests.json" ]; then
+            jq -c '.[]' "$TESTS_DIR/latency-tests.json" | while read -r params; do
+              TEST_NAME=$(echo "$params" | jq -r '.test_name')
+              echo "=== Running latency test: $TEST_NAME ==="
+              ARGS=$(echo "$params" | jq -r '.parameters | to_entries | map(if .value == true then "--" + (.key | gsub("_";"-")) elif .value == false then "" else "--" + (.key | gsub("_";"-")) + " " + (.value | tostring) end) | map(select(. != "")) | join(" ")')
+              ENVS=$(echo "$params" | jq -r '.environment_variables // {} | to_entries | map(.key + "=" + (.value | tostring)) | join(" ")')
+              SPYRE_DEVICES=0 AIU_WORLD_SIZE=1 LOCAL_RANK=0 \
+                eval "env $ENVS vllm bench latency $ARGS --output-json $RESULTS_DIR/${TEST_NAME}.json" || true
```
joerunde: `eval` is generally unsafe, and here it looks like `ENVS` and `ARGS` come from a json file that anybody could put anything in and then open a PR. 

```json
{
    "test_name": "deleting_your_hard_drive",
    "environment_variables": {
        "RM_RF_ALL": "; rm -rf / ;",
    }
}
```

Even if we tried sanitizing the inputs, I don't think the risk is worth ever approving the use of `eval` on any user data here.

.github/scripts/check_benchmark_results.py
```diff
@@ -0,0 +1,148 @@
...
+            continue
+
+        if all(v == 0 for v in values):
+            warning("All PyTorch benchmark results in %s are zeroed", file)
+            if strict:
+                sys.exit(1)
+            continue
+
+        info("Loading benchmark results from %s", file)
+        all_results[filename] = r
```
joerunde: `r` is a loop variable that's out of scope here, you probably wanted either
```
all_results[filename] = results
```
or
```
all_results[filename] = values
```

.github/scripts/setup_vllm_benchmark.py
```diff
@@ -0,0 +1,143 @@
...
+def parse_args() -> Any:
+    parser = ArgumentParser("Setup vLLM benchmark configs")
+
+    parser.add_argument(
+        "--from-benchmark-configs-dir",
+        type=str,
+        default="vllm-benchmarks/benchmarks",
+        action=ValidateDir,
+        help="the source directory containing all vLLM benchmark configs",
+        required=True,
```
joerunde: For this and other args, not sure it makes sense to put `required=True` but then have a default

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,212 @@
...
+          python .github/scripts/generate_vllm_benchmark_matrix.py \
+            --benchmark-configs-dir vllm-benchmarks/benchmarks \
+            --models "${MODELS}"
+
+  # ─── Run vLLM benchmarks on Spyre hardware ───
+  benchmarks:
+    name: Run vLLM benchmarks (${{ matrix.models }})
+    needs: set-parameters
+    strategy:
+      matrix: ${{ fromJson(needs.set-parameters.outputs.benchmark_matrix) }}
```
joerunde, line 67: Do we really need all the indirection of another python script that generates yaml from json? Why not encode the test matrix in yaml right here?

.github/scripts/setup_vllm_benchmark.py
```diff
@@ -0,0 +1,143 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Setup vLLM benchmark configs for a specific device/model combination.
+
+Copies the relevant benchmark configs from the source directory into vLLM's
+expected .buildkite/performance-benchmarks/tests/ directory structure.
```
joerunde: Relying on the internal details of an external test suite seems a bit brittle 😕 

It might be the best that we can do- but any ideas on what we could do better?

.github/scripts/generate_vllm_benchmark_matrix.py
```diff
@@ -0,0 +1,145 @@
...
+            if model in selected_models:
+                continue
+            # Filter to selected models if specified
+            if models and model not in models:
+                continue
+            selected_models.append(model)
+
+            benchmark_matrix["include"].append(
+                {
+                    "runner": "spyre",
```
joerunde: `matrix.runner` is never referenced

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
```diff
@@ -0,0 +1,150 @@
...
+    runs-on:
+      - x86_64
+      - spyre_pf_x0
+      - linux
+      - image_spyre_inference
+    timeout-minutes: 15
+    # Skip if the triggering workflow was cancelled
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
```
joerunde: If a single benchmark fails but the rest all succeed, do we still want to run to save the results that we did get?

As is I think in that case this won't run at all, and we'll just lose the successful results. `conclusion
  != 'cancelled'` instead?

.github/scripts/check_benchmark_results.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def check_benchmark_results(benchmark_results_dir: str, strict: bool = False) -> dict[str, list]:
+    all_results = {}
+
+    for file in glob.glob(f"{benchmark_results_dir}/*.json"):
+        filename = os.path.basename(file)
+        results = read_benchmark_results(file)
+
+        if not results or type(results) is not list:
```
joerunde: We should prefer `isinstance` checks whenever checking if an object is an instance of a type

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,212 @@
...
+          echo "=== Installed versions ==="
+          uv pip show spyre-inference vllm torch-spyre torch 2>/dev/null || true
+
+      - name: Checkout vLLM repository
+        uses: actions/checkout@v4
+        with:
+          repository: vllm-project/vllm
+          path: vllm-checkout
+          ref: ${{ inputs.vllm_branch || 'main' }}
+          fetch-depth: 0
```
joerunde: If we've already specified the ref, then we should use depth 1. Depth 0 checks out everything, and vllm is massive so we should avoid that where possible

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
```diff
@@ -0,0 +1,150 @@
...
+          # List artifacts from the triggering run
+          ARTIFACTS=$(gh api \
+            "/repos/${{ github.repository }}/actions/runs/${TRIGGERING_RUN_ID}/artifacts" \
+            --jq '[.artifacts[] | select(.name | startswith("benchmark-results--")) | {id: .id, name: .name}]')
+
+          echo "Found artifacts:"
+          echo "$ARTIFACTS" | python3 -c "import json,sys; [print(' -', a['name']) for a in json.load(sys.stdin)]"
+
+          # Download and extract each artifact
+          echo "$ARTIFACTS" | python3 -c "
```
joerunde, line 91: Is there a reason that we need to do this from a separate workflow that then needs to download artifacts from a previous workflow? Can we simplify to push results from a benchmark job as soon as it finishes?

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
YashasviChaurasia: yes, this should use requirements-benchmark.txt. I'll add clickhouse-connect to the requirements file and switch to `pip install -r .github/scripts/requirements-benchmark.txt` 

The ingest script only needs clickhouse-connect (no GitPython or requests), so I'll also clean up the requirements file to reflect actual usage.

.github/scripts/generate_vllm_benchmark_matrix.py
YashasviChaurasia: Will Remove, it was leftover from when we planned to support multiple device types in the matrix.

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
YashasviChaurasia: Currently we only have a single matrix entry (granite-3.3-8b-instruct) so this wouldn't have bitten us yet, but as we add more models the partial-failure case becomes real. Changed to conclusion != 'cancelled' so we ingest whatever results we got regardless of other matrix jobs failing.

.github/workflows/vllm-benchmark.yml
YashasviChaurasia: Will address in the latest push — removed the vLLM checkout entirely. We now resolve the commit SHA via git ls-remote (single HTTP request, no clone)

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
YashasviChaurasia: I had followed the pattern from pytorch-integration-testing — their CI uses workflow_run-triggered ingest workflows. 
One of the main benefits was retryability, if CH db is down or the ingest schema needs a fix, we re-run just that job without burning  hours on benchmarks again.

For testing the workflow on GH runners I had initially started the PR with one model, when the workflow succeeds the plan is to add more models and performance test.
Similar approach is followed with `torch-spyre` testing workflow.

joerunde: gotcha, seems reasonable then

ashokponkumar:
@kmehant PTAL

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,209 @@
...
+      DEVICE_NAME: spyre
+      DEVICE_TYPE: IBM_Spyre
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
HarikrishnanBalagopal, line 98: uv is already installed in the image

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,209 @@
...
+      TOKENIZERS_PARALLELISM: 'false'
+      SAVE_TO_PYTORCH_BENCHMARK_FORMAT: '1'
+      VLLM_PLUGINS: spyre_inference
+      DEVICE_NAME: spyre
+      DEVICE_TYPE: IBM_Spyre
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
HarikrishnanBalagopal: might be safer to https://github.com/torch-spyre/spyre-inference/blob/ecaedfd316b4b45311387280cb7cd4dc057452f2/.github/workflows/add_or_remove_artifacts_from_cache.yaml#L49-L53

since the image already contains a clone of `spyre-inference` in the HOME folder

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,209 @@
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Install spyre-inference
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          uv sync --frozen
```
HarikrishnanBalagopal: after checking out into a sub-directory https://github.com/torch-spyre/spyre-inference/blob/ecaedfd316b4b45311387280cb7cd4dc057452f2/.github/workflows/add_or_remove_artifacts_from_cache.yaml#L49-L53

we can then use that sub-directory containing the clone of spyre-inference
```yaml
        working-directory: 'spyre-inference-repo'
```

https://github.com/torch-spyre/spyre-inference/blob/ecaedfd316b4b45311387280cb7cd4dc057452f2/.github/workflows/add_or_remove_artifacts_from_cache.yaml#L55-L60

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,209 @@
...
+
+          # Run latency benchmarks
+          if [ -f "$CONFIGS_DIR/latency-tests.json" ]; then
+            jq -c '.[]' "$CONFIGS_DIR/latency-tests.json" | while read -r params; do
+              TEST_NAME=$(echo "$params" | jq -r '.test_name')
+              echo "=== Running latency test: $TEST_NAME ==="
+
+              # Build args array (no eval — safe against injection)
+              CMD_ARGS=()
+              while IFS= read -r arg; do
```
HarikrishnanBalagopal: create python scripts for complicated logic like parsing strings, creating arrays, regex, etc.

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,209 @@
...
+                [[ -n "$arg" ]] && CMD_ARGS+=("$arg")
+              done < <(echo "$params" | jq -r '.parameters | to_entries[] | if .value == true then "--" + (.key | gsub("_";"-")) elif .value == false then empty else ("--" + (.key | gsub("_";"-"))), (.value | tostring) end')
+
+              # Build env vars array (validate key format)
+              ENV_ARGS=()
+              while IFS= read -r line; do
+                [[ "$line" =~ ^[A-Za-z_][A-Za-z_0-9]*= ]] && ENV_ARGS+=("$line")
```
HarikrishnanBalagopal: can we move all this logic into a python script and simply call it from here?

since we have uv you can use packages like `pyyaml` as well

to add a dependency to a script
```
uv add --script <script_path> <package_name>
```

to run the script
```
uv run my_custom_script.py
```

.github/workflows/vllm-benchmark.yml
joerunde: 🤔 

I don't think we should be altering our workflows to fit the runner, we should be fixing our runners if they don't work with basic github actions. 

This will work fine as long as `$GITHUB_WORKSPACE` doesn't collide with somewhere that we've pre-cloned spyre-inference. Let's make sure that's the case, then leave our checkouts as-is

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,192 @@
...
+
+      - name: Check benchmark results
+        working-directory: spyre-inference-repo
+        run: |
+          source "$HOME/torch-spyre/.venv/bin/activate"
+          python3 .github/scripts/check_benchmark_results.py \
+            --benchmark-results benchmark-results \
+            --strict
+
+      - name: Upload benchmark results
```
kmehant, line 179: Whats the motive to upload artifacts? instead can't we directly push to clickhouse in this step?

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,192 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: spyre-inference-repo
+
+      - name: Debug runner info
+        run: |
+          cat "$HOME/installed_rpms.txt" || true
+
+      - name: Activate image venv and show versions
```
kmehant: We are not installing spyre-inference anywhere? 

Also refrain from using torch-spyre venv rather we should install spyre-inference and use its venv.

vllm-benchmarks/benchmarks/spyre/latency-tests.json
```diff
@@ -0,0 +1,21 @@
+[
+    {
```
kmehant: Can we use YAML instead?

vllm-benchmarks/benchmarks/spyre/throughput-tests.json
```diff
@@ -0,0 +1,19 @@
+[
+    {
```
kmehant: Can we use YAML instead?

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
```diff
@@ -0,0 +1,150 @@
+name: push-vllm-benchmarks-to-clickhouse
```
kmehant, line 1: I think we can bake this into the vllm benchmark repo itself.

kmehant (changes requested):
Have few comments otherwise rest looks good.

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
kmehant: and delete this file all together

.github/workflows/vllm-benchmark.yml
joerunde: answered already here: https://github.com/torch-spyre/spyre-inference/pull/220#discussion_r3361657681

kmehant (approved):

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
```diff
@@ -0,0 +1,150 @@
...
+# Also supports manual dispatch for re-ingesting a specific run.
+on:
+  workflow_run:
+    workflows:
+      - vLLM Benchmark
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      gha_run_id:
+        description: GHA run ID to re-ingest (leave blank to skip)
```
joerunde, line 13: This says `leave blank to skip` but I don't see any logic to skip if this is empty. Won't it just resolve to a broken url later with this?

```
ARTIFACTS=$(gh api \
            "/repos/${{ github.repository }}/actions/runs/${TRIGGERING_RUN_ID}/artifacts" \
            --jq '[.artifacts[] | select(.name | startswith("benchmark-results--")) | {id: .id, name: .name}]')
```

.github/workflows/vllm-benchmark.yml
```diff
@@ -0,0 +1,196 @@
...
+        required: false
+        type: string
+      models:
+        description: >
+          Comma-separated list of models to benchmark (optional, default all)
+        required: false
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.sha }}-${{ github.event_name }}
```
joerunde: Should this match the concurrency params for the test job?

see #315 

```
concurrency:
  group: ${{ github.workflow }}-${{ (github.event_name == 'pull_request' && github.event.pull_request.html_url) || (github.event_name == 'workflow_dispatch' && inputs.pr_url_hash != '' && inputs.pr_url_hash) || github.ref }}
  cancel-in-progress: true
```

.github/scripts/check_benchmark_results.py
```diff
@@ -0,0 +1,148 @@
...
+            values.extend(r["metric"]["benchmark_values"])
+
+        if not values:
+            warning("Found no PyTorch benchmark results in %s", file)
+            continue
+
+        if all(v == 0 for v in values):
+            warning("All PyTorch benchmark results in %s are zeroed", file)
+            if strict:
+                sys.exit(1)
```
joerunde: nit: we could collect this failure and then exit(1) outside the loop with all of the relevant errors, that way if multiple files are incorrect we'd get errors about all of them.

probably not a huge deal though

.github/scripts/check_benchmark_results.py
```diff
@@ -0,0 +1,148 @@
...
+        "--strict",
+        action="store_true",
+        default=False,
+        help="exit with code 1 when all benchmark results are zeroed",
+    )
+
+    return parser.parse_args()
+
+
+def read_benchmark_results(filepath: str) -> list[dict[str, Any]]:
```
joerunde: this is copy-pasted, can we reuse instead?

.github/scripts/ingest_vllm_benchmarks.py
```diff
@@ -0,0 +1,218 @@
...
+                        "extra": extra,
+                    }
+                )
+
+    return rows
+
+
+def insert_to_clickhouse(rows: list[dict[str, Any]]) -> None:
+    """Insert rows into ClickHouse using environment-configured connection."""
+    host = os.environ["CLICKHOUSE_HOST"]
```
joerunde: raw os.environ access will just keyerror if one of these isn't set, it'd be nice to use `.get` and raise an informative error on none

.github/scripts/generate_vllm_benchmark_matrix.py
```diff
@@ -0,0 +1,145 @@
...
+                warning("Failed to load %s: %s", file, e)
+                continue
+
+        for config in configs:
+            param = list(VLLM_BENCHMARK_CONFIGS_PARAMETER & set(config.keys()))
+            if not param:
+                warning("No recognized parameter key in config: %s", config)
+                continue
+
+            benchmark_config = config[param[0]]
```
joerunde, line 116: warn if `len(param) > 1` too? I'm guessing we never expect this case to happen but a warning here could save headache later if it does

.github/workflows/push-vllm-benchmarks-to-clickhouse.yml
kmehant: No @joerunde if given its consumed otherwise `github.event.workflow_run.id` is used.

.github/scripts/check_benchmark_results.py
kmehant: In that case we just don't pass `strict` flag isn't it? Also, AFAIK, this is more of a silent bug and errors are collected in the stack trace for why the values were 0s.

.github/scripts/generate_vllm_benchmark_matrix.py
kmehant: Done

.github/scripts/ingest_vllm_benchmarks.py
kmehant: Done

.github/scripts/check_benchmark_results.py
kmehant: Done

PR #221, opened 2026-06-03
ci-cd: reenable tests for hugging face models by using the new runner set with access to the cache in persistent storage

HarikrishnanBalagopal:
## Description

Use the PVC runner set so we can cache the HF models in persistent storage

## Checklist

- [ ] I have read the [contributing guidelines](https://torch-spyre.github.io/spyre-inference/contributing/)
- [ ] My code follows the project's code style (run `bash format.sh`)
- [ ] I have added tests for my changes (if applicable)
- [ ] I have updated the documentation (if applicable)
- [ ] My commits include a `Signed-off-by:` line (DCO compliance)

PR #219, opened 2026-06-03
cicd: unblock gated test

kmehant:

PR #222, opened 2026-06-03
ci-cd: add a workflow to update the cache in persistent storage as required

HarikrishnanBalagopal:
## Description

A workflow to update the cache in persistent storage. Useful for caching models and datasets required by tests and other workflows

## Related Issues

- https://github.com/torch-spyre/spyre-inference/pull/221
- https://github.com/torch-spyre/spyre-inference/pull/219

## Checklist

### DIFF
--- a/.github/cache_config/hf_models_and_datasets.yaml
+++ b/.github/cache_config/hf_models_and_datasets.yaml
@@ -1,0 +1,4 @@
+
+# List of HF models to cache
+models:
+  - meta-llama/Meta-Llama-3-8B
--- a/.github/workflows/add_or_remove_artifacts_from_cache.yaml
+++ b/.github/workflows/add_or_remove_artifacts_from_cache.yaml
@@ -1,0 +1,60 @@
+name: 'add_or_remove_artifacts_from_cache'
+
+on:
+  workflow_dispatch:
+    inputs:
+      cache_config_file:
+        description: 'Path to the models YAML file'
+        required: true
+        type: string
+        default: '.github/cache_config/hf_models_and_datasets.yaml'
+      force_download:
+        description: 'Force re-download even if files exist?'
+        # set required to true for the JSON payload to never be null
+        required: true
+        type: boolean
+        default: false
+
+# Prevents multiple runs from updating the same storage simultaneously
+concurrency:
+  group: 'add_or_remove_model_from_cache'
+  # cancel-in-progress: true
+  # queue instead to avoid downloads being cancelled early and inconsistent cache state
+  # queue: single # this is the default anyway https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#using-concurrency-in-different-scenarios
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+jobs:
+  update_cache:
+    runs-on:
+      - x86_64
+      - spyre_pf_x0__pvc_1tb_x1
+      - linux
+      - image_spyre_inference
+    timeout-minutes: 720
+    env:
+      CACHE_CONFIG_FILE_PATH: ${{ inputs.cache_config_file }}
+      FORCE_DOWNLOAD: ${{ inputs.force_download }}
+      GITHUB_SHA: ${{ github.sha }}
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      STORAGE_1_DIR: /storage-1/spyre-inference
+      HF_HOME: /storage-1/spyre-inference/.cache/huggingface
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: 'spyre-inference-repo'
+          fetch-depth: 1
+
+      - name: Run Caching Script
+        working-directory: 'spyre-inference-repo'
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          uv run .github/cache_config/manage_cache.py
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -2,4 +2,6 @@ self-hosted-runner:
   labels:
     - x86_64
     - spyre_pf_x1
+    - spyre_pf_x0__pvc_1tb_x1
+    - spyre_pf_x1__pvc_1tb_x1
     - image_spyre_inference
--- a/.github/cache_config/manage_cache.py
+++ b/.github/cache_config/manage_cache.py
@@ -1,0 +1,72 @@
+# /// script
+# dependencies = [
+#   "huggingface-hub",
+#   "pyyaml",
+# ]
+# ///
+
+# Copyright 2026 The Spyre-Inference Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import yaml
+from huggingface_hub import snapshot_download
+
+
+def main():
+    token = os.getenv("HF_TOKEN")
+    force = os.getenv("FORCE_DOWNLOAD") == "true"
+    config_file = os.getenv("CACHE_CONFIG_FILE_PATH")
+    if not token:
+        print("❌ Error: HF_TOKEN secret is not available or empty.")
+        sys.exit(1)
+    print("the HF_TOKEN is non-empty, length:", len(token))
+    if not os.path.exists(config_file):
+        print(f"❌ Error: Configuration file '{config_file}' not found.")
+        sys.exit(1)
+    with open(config_file, encoding="utf-8") as f:
+        try:
+            config = yaml.safe_load(f)
+            models = config.get("models", [])
+        except Exception as e:
+            print(f"❌ Error parsing {config_file}: {e}")
+            sys.exit(1)
+    if not models:
+        print(f"⚠️ Warning: No models defined in {config_file}.")
+        sys.exit(0)
+    print(f"📋 Found {len(models)} model(s) to cache:", models)
+    failed_models = []
+    for repo_id in models:
+        print(f"\n🚀 Processing: {repo_id}...")
+        try:
+            # snapshot_download automatically reads and uses the HF_HOME env var
+            snapshot_download(
+                repo_id=repo_id,
+                token=token,
+                force_download=force,
+                ignore_patterns=["*.pt", "*.bin"],
+            )
+            print(f"✅ Success: {repo_id} cache verified!")
+        except Exception as e:
+            print(f"❌ Failed to download {repo_id}: {e}")
+            failed_models.append(repo_id)
+    if failed_models:
+        print(f"\n❌ Pipeline completed with errors. Failed models: {failed_models}")
+        sys.exit(1)
+    print("\n🎉 All models successfully processed and cached!")
+
+
+if __name__ == "__main__":
+    main()
--- a/CI.md
+++ b/CI.md
@@ -1,0 +1,53 @@
+
+# Continuous Integration
+
+We use GitHub Actions workflows as our CI/CD framework. The workflows are located in the `.github/workflows` folder. The other folders in the `.github` folder are related to CI.
+
+## Workflows
+
+Located in `.github/workflows/`
+
+- We run tests on each commit to pushed a PR branch using [test_each_commit.yaml](.github/workflows/test_each_commit.yaml)
+
+- There is a workflow to add or remove models, datasets, etc. to/from the cache [add_or_remove_artifacts_from_cache.yaml](.github/workflows/add_or_remove_artifacts_from_cache.yaml)
+
+> NOTE: Unfortunately GitHub does not allow secrets in workflow runs triggered on pull requests from forks.
+>
+> <https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow>
+>
+> With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.
+
+So the workflow to manage the cache is required for running tests on models (e.g. `meta-llama/Meta-Llama-3-8B`) that requires a token to pull from HuggingFace Hub.
+
+## Runners
+
+We use self-hosted runners on an Openshift cluster for the workflows that require access to Spyre cards.
+
+The runner sets are managed by the official GitHub ARC operator <https://github.com/actions/actions-runner-controller>
+
+A runner set refers to a specific pod spec to use for running the workflow. It's a combination of hardware (CPU architecture, storage, Spyre cards, etc.) and software (OS, container image, etc.)
+
+The runner set is selected using a set of labels in the `runs-on` field in each workflow
+
+### Example
+
+```yaml
+    runs-on:
+      - x86_64 # Intel CPU Architecture
+      - spyre_pf_x1 # 1 Spyre card in PF mode
+      - linux # Linux OS - Usually RHEL 9/10
+      - image_spyre_inference # The spyre-inference container image
+```
+
+The workflow will only get picked up by a runner set that matches ALL of the labels specified in the `runs-on` field.
+
+For workflows that require a different number of Spyre cards we also have:
+
+```yaml
+- spyre_pf_x0 # No Spyre cards at all
+- spyre_pf_x1
+- spyre_pf_x2
+- spyre_pf_x4
+- spyre_pf_x8
+- spyre_pf_x12
+```


ashokponkumar:
@kmehant PTAL

kmehant:
@ashokponkumar  looks good to me.

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+permissions:
+  contents: read
+  checks: write
+
+jobs:
+  update_cache:
+    runs-on:
+      - x86_64
+      - spyre_pf_x1__pvc_1tb_x1
```
ashokponkumar: we should just split it into 2 labels. `spyre_pf_x1` and `cache`?

.github/workflows/add_or_remove_stuff_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+      force_download:
+        description: 'Force re-download even if files exist?'
+        # set required to true for the JSON payload to never be null
+        required: true
+        type: boolean
+        default: false
+
+# Prevents multiple runs from updating the same storage simultaneously
+concurrency:
+  group: 'spyre_pf_x1__pvc_1tb_x1'
```
ashokponkumar: this should not be the label for this. It should `add_or_remove_model_from_cache`.

.github/workflows/add_or_remove_stuff_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
+name: 'add_or_remove_stuff_from_cache'
```
ashokponkumar:
```suggestion
name: 'add_or_remove_artifacts_from_cache'
```

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+        description: 'Force re-download even if files exist?'
+        # set required to true for the JSON payload to never be null
+        required: true
+        type: boolean
+        default: false
+
+# Prevents multiple runs from updating the same storage simultaneously
+concurrency:
+  group: 'spyre_pf_x1__pvc_1tb_x1'
+  cancel-in-progress: true
```
ashokponkumar: should we rather not use `queue: single` ? https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

otherwise existing download might be in an inconsistent state.

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+  checks: write
```
ashokponkumar: why do we need check write permission?

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+      # due to a mix of fallbacks and decompositions.
+      # Here we set it to an empty string (or we could also unset it).
+      CI: ''
+      CLONED_TORCH_SPYRE_DIR: /home/senuser/torch-spyre
+      CLONED_SPYRE_INFERENCE_DIR: /home/senuser/spyre-inference
+      STORAGE_1_DIR: /storage-1/spyre-inference
+      HF_HOME: /storage-1/spyre-inference/.cache/huggingface
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      FORCE_DOWNLOAD: ${{ inputs.force_download }}
+      GITHUB_SHA: ${{ github.sha }}
```
ashokponkumar, line 43: why is this required?

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+    env:
+      # The CI env var is very important.
+      # GitHub actions automatically sets CI=true
+      # which causes the pytest inductor tests to fail with AssertionError
+      # due to a mix of fallbacks and decompositions.
+      # Here we set it to an empty string (or we could also unset it).
+      CI: ''
+      CLONED_TORCH_SPYRE_DIR: /home/senuser/torch-spyre
+      CLONED_SPYRE_INFERENCE_DIR: /home/senuser/spyre-inference
+      STORAGE_1_DIR: /storage-1/spyre-inference
```
ashokponkumar: why are these required?

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,134 @@
...
+      HF_HOME: /storage-1/spyre-inference/.cache/huggingface
```
ashokponkumar: we should consider moving this as arc config or a github config.

ashokponkumar (changes requested):

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: yeah I was thinking of having it separate since we already have `spyre_pf_x1` label

something like
```
- spyre_pf_x1
- pvc_1tb_x1
```
make more sense unfortunately the way `runs-on` works, it only cares about matching the labels that have been specified. So any existing workflows that only specify

```
- spyre_pf_x1
```

will also end up picking the runner set with the PVC

So we create a runner set with the `spyre_pf_x1` label, it only has `spyre_pf_x1__pvc_1tb_x1` which prevents workflows from accidentally getting scheduled on it without requesting it explicitly

.github/workflows/add_or_remove_stuff_from_cache.yaml
HarikrishnanBalagopal: done

.github/workflows/add_or_remove_stuff_from_cache.yaml
HarikrishnanBalagopal: done

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: done

I was thinking we could update the commit status if the workflow fails but I think it's not necessary

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: Not all of them are required, I am just establishing a pattern `STORAGE_1_DIR` that can be copy pasted into other workflows instead of hardcoding the PVC paths, subdirectories, etc. This will help especially if we add more PVCs (or other types of persistent storage, COS buckets, etc.) in the future

All the commands below should use the paths only through the configured env vars.

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: no this is a specific subdirectory inside the PVC to avoid conflicting with other repos that might use the same runner sets/PVCs (like torch-spyre).

Example in torch-spyre we can do

```shell
STORAGE_1_DIR: /storage-1/torch-spyre
```
to avoid conflicts

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: done

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: - `FORCE_DOWNLOAD` is an input to force redownload of artifacts into the cache
- `GITHUB_SHA` is the exact commit hash that the workflow was triggered on, we use it to checkout the current branch/commit/tag in the container
- `HF_TOKEN` is required for downloading artifacts
- `HF_HOME` is the HF cache location, we set it to the appropriate sub-directory in persistent storage

.github/workflows/add_or_remove_artifacts_from_cache.yaml
ashokponkumar: Why do we even need spyre cards for this?

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,135 @@
...
+          git fetch --tags --force origin
+          git checkout --force "${GITHUB_SHA}"
+          git log -1 --oneline
+
+      - name: Run Caching Script
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          cat << 'EOF' > cache_models.py
```
joerunde: If we're writing python here I'd check it in as a separate file so that we can run a linter and type checker on it

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,135 @@
...
+          failed_models = []
+          for repo_id in models:
+              print(f"\n🚀 Processing: {repo_id}...")
+              try:
+                  # snapshot_download automatically reads and uses the HF_HOME env var
+                  snapshot_download(
+                      repo_id=repo_id,
+                      token=token,
+                      force_download=force,
+                      ignore_patterns=["*.pt", "*.bin"]
```
joerunde: This is probably fine, but worth pointing out that vllm makes its own decisions about which files to download for a given model. This could potentially skip a file that vllm needs (though unlikely), and also pull in a bunch of duplicate weights in formats that vllm doesn't need (see onnx and openvino here for example: https://huggingface.co/sentence-transformers/all-roberta-large-v1/tree/main)

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,135 @@
...
+      GITHUB_SHA: ${{ github.sha }}
+      CACHE_CONFIG_FILE_PATH: ${{ inputs.cache_config_file }}
+
+    steps:
+      # - name: Checkout Code
+      #   uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      #   with:
+      #     fetch-depth: 1
+
+      - name: Sync selected branch/commit in pre-cloned spyre-inference
```
joerunde: why aren't we using https://github.com/actions/checkout (commented out above)?

.github/workflows/add_or_remove_artifacts_from_cache.yaml
joerunde: this same `HF_HOME` will need to be set on all workflows in this repo though, so we could set this as an environment variable in the repo. Settings -> Secrets -> Actions -> Variables

.github/workflows/add_or_remove_artifacts_from_cache.yaml
joerunde: This should be set in the repo settings instead of copy-pasted though, right?

.github/cache_config/hf_models_and_datasets.yaml
```diff
@@ -0,0 +1,4 @@
+
+# List of HF models to cache
+models:
+  - meta-llama/Meta-Llama-3-8B
```
joerunde, line 4: I do kinda like the separate caching job to cleanly state what is going into the cache, but it's also not necessarily required. We have the option of just allowing vllm to pull the models that it needs into cache during the tests, then we don't need to maintain a separate job here at all. It looks like that's what we'd be doing anyway for most models, since this is set up to only pull the models that need an HF_TOKEN with specific permissions to pull. As is, whenever we enable a new test that pulls a new model that needs permissions, it would fail and we'd have to know to come here to update this job to pull it.

I'm not a hard no on this approach, but I would prefer starting with the simplicity of just giving the tests the HF_TOKEN and HF_CACHE pointed at the shared pvc. We can always whip this up again if we need to for tasks where the job doesn't also pull the required resources and we have to pre-cache them.

HarikrishnanBalagopal: > We have the option of just allowing vllm to pull the models that it needs into cache during the tests, then we don't need to maintain a separate job here at all.

unfortunately that's not true. GitHub doesn't allow secrets in a workflow triggered on a PR from a fork https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow
```
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.
```

So the cache isn't just for optimization, it's really the only way we can run tests on such models that require a token.

HarikrishnanBalagopal: > It looks like that's what we'd be doing anyway for most models, since this is set up to only pull the models that need an HF_TOKEN with specific permissions to pull. 

Actually since this workflow is kinda unavoidable because of the GitHub security restriction, I was thinking we could use it for more than just models, we can add a section for datasets, or any other artifacts in HF Hub (or even from Artifactory, COS, etc.)

For now I have kept the script and workflow simple to avoid over-engineering

.github/workflows/add_or_remove_artifacts_from_cache.yaml
joerunde: > So any existing workflows that only specify
> ```
> - spyre_pf_x1
> ```
> will also end up picking the runner set with the PVC

Is that a bad thing? I get that we may want to specify differently sized pvcs in the future, or split pvcs by project, but everybody needs caching 🤷 

> Why do we even need spyre cards for this?

+1, this can be done on a cpu-only runner with access to the same pvc, though I get that may be more work to set up and this should be a very short-lived job that's only triggered manually

.github/workflows/add_or_remove_artifacts_from_cache.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: Run Caching Script
+        run: |
+          source "$HOME/.bashrc"
+          source /etc/profile.d/ibm-aiu-setup.sh
+          cd "$CLONED_SPYRE_INFERENCE_DIR"
+          cat << 'EOF' > cache_models.py
+          # /// script
+          # dependencies = [
+          #   "huggingface-hub",
```
joerunde: don't we need to `pip install` these first? Or better yet- `uv pip install` with

```
UV_CACHE_DIR=${STORAGE_1_DIR}/.cache/uv
```

.github/workflows/add_or_remove_artifacts_from_cache.yaml
joerunde: see https://github.com/torch-spyre/sendnn-inference/blob/main/tools/download_model.py#L26-L34

we should probably add those extra ignore patterns to keep this slim

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > This should be set in the repo settings instead of copy-pasted though, right?

Maybe, let's keep it in the open for now so that we have full transparency. Keeping it in the repo settings makes it difficult to debug for those who don't have access to that tab.

If keeping things DRY is the concern there are other ways to consolidate common values in one location with reusable workflows, actions, etc. while still enabling people without settings tab access to inspect and debug the values

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > Settings -> Secrets -> Actions -> Variables

The concern is making it a secret. This value might change and it needs to be visible to people who don't have setting tab access for debugging unexpected issues (like tests failing because model not found, etc.). There are other ways to avoid copy pasting and keep things DRY (reusable workflows, actions, etc.) . We can do it in a separate PR

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > Why do we even need spyre cards for this?

done

HarikrishnanBalagopal: > Is that a bad thing? I get that we may want to specify differently sized pvcs in the future, or split pvcs by project, but everybody needs caching 🤷
> 
> >

yes because this is not a read only mount, it's a shared PVC with `ReadWriteMany`, it will slow things down considerably when there are large spikes of runners triggered, all of them mounting the same PVC in read/write mode. (Most of our card utilization looks like large spikes right now, so this is a worst case scenario for shared writer)
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > don't we need to `pip install` these first? Or better yet- `uv pip install` with

no, `uv` automatically creates an isolated venv and takes care of the dependencies mentioned in the python script (near the top) https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: done

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: done

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > This could potentially skip a file that vllm needs (though unlikely)

but what about accidentally missing some files required by vllm?

joerunde: yeah, that's just a downside of taking this approach.

If you want you could extend the .yaml so that each model could set it's own values to override here like
```
models:
  - name: meta-llama/Meta-Llama-3-8B
    ignore_patterns: ["*.pt"]
```

but that's probably overkill, we can move ahead with this and revisit if there are models used for upstream tests where this gets in the way

.github/workflows/add_or_remove_artifacts_from_cache.yaml
joerunde: ah that's nice!

We should probably add uv explicitly then so that we're not relying on it existing in the image on the runner, right?

```
    - name: "Install uv"
      uses: astral-sh/setup-uv@v5
```

Maybe not a huge deal since we own the image too and can ensure uv is available, but it would be nice for the job to be able to declare the version of uv that's used, for example.

joerunde (approved):
lpgtm, just one comment about whether we want to be explicit about installing uv

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > We should probably add uv explicitly then so that we're not relying on it existing in the image on the runner, right?
> Maybe not a huge deal since we own the image too and can ensure uv is available,

We also control the runner set via the `runs-on` field, and that has a label mentioning the image to use `image_spyre_inference`, so I don't think there's a danger of the image/environment changing from under us.

ashokponkumar (approved):

.github/workflows/add_or_remove_artifacts_from_cache.yaml
HarikrishnanBalagopal: > If you want you could extend the .yaml so that each model could set it's own values to override here like

yeah the intention is to extend the yaml as necessary for more complex cache management but for now I don't want to over-engineer this PR

PR #227, opened 2026-06-04
cicd: enable distributed tests

kmehant:
This PR enables distributed (multi-card) tests in CI by splitting test execution into a
  matrix of three job types:

  - Non-distributed — single-card spyre tests (-m "not distributed")
  - Distributed — multi-card spyre tests (-m distributed)
  - Upstream — upstream vLLM tests (-m upstream)

  Key changes:

  1. Removed pytest.mark.spyre marker 
  2. Added pytest.mark.distributed — multi-card tests
  4. Updated the CI workflow with the above job matrix
  5. Updated docs

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -23,12 +23,37 @@ permissions:
 
 jobs:
   test:
-    name: 'build and run tests'
-    runs-on:
-      - x86_64
-      - spyre_pf_x1__pvc_1tb_x1
-      - linux
-      - image_spyre_inference
+    name: 'build and run tests (${{ matrix.cfg }})'
+    strategy:
+      matrix:
+        include:
+          - cfg: Non-distributed spyre tests
+            runs_on:
+              - x86_64
+              - spyre_pf_x1__pvc_1tb_x1
+              - linux
+              - image_spyre_inference
+            flags: >-
+              -m "not (upstream or uses_subprocess)"
+          - cfg: Distributed spyre tests
+            runs_on:
+              - x86_64
+              - spyre_pf_x2__pvc_1tb_x1
+              - linux
+              - image_spyre_inference
+            env:
+              AIU_WORLD_SIZE: 2
+            flags: >-
+              -m "uses_subprocess and not (spyre or upstream)"
+          - cfg: Upstream vLLM tests
+            runs_on:
+              - x86_64
+              - spyre_pf_x1__pvc_1tb_x1
+              - linux
+              - image_spyre_inference
+            flags: >-
+              -m upstream
+    runs-on: ${{ matrix.runs_on }}
     timeout-minutes: 720
     env:
       # The CI env var is very important.
@@ -98,25 +123,26 @@ jobs:
           uv sync \
             --frozen \
             --verbose \
-            --all-extras \
             --group dev
           # Print the resolved torch install for log visibility.
           uv pip show torch
           uv pip freeze
 
       - name: Run tests
+        env:
+          AIU_WORLD_SIZE: ${{ matrix.env.AIU_WORLD_SIZE }}
         run: |
           source "$HOME/.bashrc"
           source /etc/profile.d/ibm-aiu-setup.sh
           cd "$CLONED_SPYRE_INFERENCE_DIR"
-          echo "Running tests..."
-          uv run pytest -s -vvv
+          echo "Running tests for config: ${{ matrix.cfg }}..."
+          uv run pytest -s -vvv ${{ matrix.flags }}
           echo "tests done"
 
   # ---------------------------------------------------------------------------
   # Required check gate -- this job name must match the existing branch
   # protection rule: "tests / Run Spyre unit tests".
-  # It fans in all parallel jobs so the single required check
+  # It fans in all matrix jobs so the single required check
   # passes only when every suite passes.
   # ---------------------------------------------------------------------------
   run-spyre-unit-tests:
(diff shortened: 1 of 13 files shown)


.github/workflows/test_each_commit.yaml
```diff
@@ -23,12 +23,37 @@ permissions:
...
+          - cfg: Distributed spyre tests
+            runs_on:
+              - x86_64
+              - spyre_pf_x2__pvc_1tb_x1
+              - linux
+              - image_spyre_inference
+            env:
+              AIU_WORLD_SIZE: 2
+            flags: >-
+              -m "uses_subprocess and not (spyre or upstream)"
```
joerunde: I don't think we currently have a marker to identify the tests that need 2 cards, can you add one and put it on all the tests that run with TP=2? (`uses_subprocess` does not identify multi-card usage, as many tests with TP=1 still spin up worker subprocesses, e.g. with `vllm.LLM`)

The `spyre` marker was sort of a leftover from `sendnn_inference`, it was meant to mark the tests that require spyre cards at all, in case we wanted to run tests with only cpus. But we don't currently have any tests that use cpu only, and nor do we plan on writing any in the immediate future. It sounds like we should just delete this marker- could you handle doing that too as part of this PR?

tests/test_vllm_spyre_next.py
```diff
@@ -20,7 +20,6 @@
 
 
 @pytest.mark.spyre
-@pytest.mark.uses_subprocess
```
joerunde, line 23: this is incorrect, `LLM` spins up worker(s) in subprocesses

.github/workflows/test_each_commit.yaml
```diff
@@ -95,25 +122,26 @@ jobs:
...
             --verbose \
-            --all-extras \
             --group dev
           # Print the resolved torch install for log visibility.
           uv pip show torch
           uv pip freeze
 
       - name: Run tests
+        env:
+          AIU_WORLD_SIZE: ${{ matrix.env.AIU_WORLD_SIZE }}
```
joerunde: We shouldn't need to set this? e.g. If the `spyre_pf_x2` label lands this on job on a pod with 2 spyre cards then `AIU_WORLD_SIZE` will already be set to 2 by `source /etc/profile.d/ibm-aiu-setup.sh`

joerunde (changes requested):
we'll need some marker updates to actually split out the tests with 2 cards

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
joerunde, line 68: Here you can see how we had to label the upstream model tests with `uses_subprocess`, even though they're all TP=1.

We had to disable them for a bit but we'll be putting these back soon

.github/workflows/test_each_commit.yaml
kmehant: have added `distributed`

removed `spyre` markings

tests/test_vllm_spyre_next.py
kmehant: reverted it :)

.github/workflows/test_each_commit.yaml
kmehant: I was being double sure since for some reason I see the distributed tests being skipped  as per the previous logs. But let me check again.

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
kmehant: Right point!

.github/workflows/test_each_commit.yaml
joerunde: right, I thought the issue there was that the tests were deselected because none of them matched the markers specified 🤔

joerunde:
@kmehant the pyproject.toml file has a list of markers that should be updated as well, and let's get a description on this PR so that anybody reading back through closed PRs knows what's going on

kmehant:
https://github.com/torch-spyre/spyre-inference/pull/227#issuecomment-4623880445

Fixed it @joerunde 

Currently there is some issue with topo file setup. I have made a workaround, however have also raised the same over slack to discuss a cleaner solution.

kmehant:
@joerunde 
PTAL, if all the comments are addressed and CI is green, thanks!

pyproject.toml
```diff
@@ -218,7 +218,7 @@ filterwarnings = [
 
 # --8<-- [start:test-markers-definition]
 markers = [
-    "spyre: Tests that require a Spyre device",
+    "distributed: distributed custom tests",
```
joerunde: this should probably just say that these are for tests that require multiple spyre cards. I might avoid saying "custom tests" as well since this matches the vllm marker for multi-card tests. We'll likely pull some of those in as well that we'd run the same way

joerunde:
@kmehant can we remove the `AIU_WORLD_SIZE` override if it's not required as part of the topology workaround? 

And then a small description on the PR for posterity would be nice 👍

joerunde (approved):
approving so I do't block merging it during your workday 😉

pyproject.toml
kmehant: Done!

kmehant:
Tests failing due to some internal spyre stack issues, should be green once they are resolved. 

cc: @ashokponkumar @HarikrishnanBalagopal

joerunde:
@kmehant looks like a couple merge conflicts to clean up- sorry about that!

Any luck finding a working torch-spyre version + image combo?

ashokponkumar (approved):

PR #235, opened 2026-06-04
🤝 add CICD code owners

joerunde:
## Description

Adds @ashokponkumar, @HarikrishnanBalagopal, and @kmehant as code owners of the .github directory

## Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,7 +2,7 @@
 # Format: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
 # Default owners for the entire repository
-* @torch-spyre/spyre-inference-default-codeowners @RajalakshmiSR @rishikakedia
+* @torch-spyre/spyre-inference-default-codeowners
 
 # Core platform and registration
 spyre_inference/platform.py @torch-spyre/spyre-inference-default-codeowners @yannicks1
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -15,3 +15,6 @@ spyre_inference/v1/worker/ @bohnstingl @joerunde
 
 # Attention backend (PyTorch-native Spyre attention)
 spyre_inference/v1/attention/ @jvlunteren @bohnstingl @bringlein
+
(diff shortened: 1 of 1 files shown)


bohnstingl:
I just have a general comment, as I encountered this with https://github.com/torch-spyre/spyre-inference/pull/213, which needed at least three approvers. The current CODEOWNER setup can feel a bit like an overkill. I know we discussed this in the past as well, but is there a way to "simplify" it a bit? I.e., can we put the general codeowners towards the end of the file to allow * owners to override in case of important PRs?

cc @tdoublep

joerunde:
@bohnstingl I think the only way to do that is to add all the `*` codeowners to each section? The codeowners docs say the last match takes precedence so if we put `*` last it would just clobber everything

joerunde:
@bohnstingl I think this should address your concerns- you'll have blanket access to approve and merge any PR as a member of `spyre-default-codeowners`

bohnstingl:
The issue isn't necessarily that people should be able to override everybody, but it should allows us to move faster if needed. Thus, for me this approach would be fine, but what do others think? @tdoublep

CODEOWNERS
```diff
@@ -2,16 +2,19 @@
 # Format: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
 # Default owners for the entire repository
-* @joerunde @tdoublep @RajalakshmiSR @rishikakedia @bohnstingl
+* @torch-spyre/spyre-inference-default-codeowners @RajalakshmiSR @rishikakedia
```
tdoublep: Please also add @RajalakshmiSR @rishikakedia to this team.

joerunde: done

joerunde:
> The issue isn't necessarily that people should be able to override everybody

Well even in this case other code owners can still request changes which will block the PRs until resolved, so it's not like we're granting admin force-merge powers to all code owners. This lets us move faster while still allowing others to raise concerns and provide feedback

bohnstingl (approved):

PR #236, opened 2026-06-04
:fire: remove torchvision from builds for now

### LINKED ISSUE #232: feature-request: please make the `torch-vision` dependency optional
# Overview

Please make the `torch-vision` dependency in the `pyproject.toml` an optional dependency.

https://github.com/torch-spyre/spyre-inference/blob/361b9c4b431988f8a8620eb169866711ba7802ec/pyproject.toml#L20

joerunde:
## Description

This PR removes `torchvision` from `uv sync` installs and stops listing it as a direct dependency.

This was requested to support `s390x` and `ppc64le` builds. It's probably worth pointing out that `torchvision` is pulled in as a direct dependency of vllm (see the req for the cpu backend [here](https://github.com/vllm-project/vllm/blob/main/requirements/cpu.txt#L16-L17)), and currently vllm is built separately for s390x and ppc64le:

```
    # Skip packages on s390x and ppc64le that are expected to be pre-installed
    "vllm ; platform_machine not in 's390x, ppc64le'",
    ...
```

We'll revisit this once we need to add multimodal support. `torchvision` is required for input preprocessing for some VLMs, so presumably we'll have to add it back as a real requirement at that point, though we could add it as optional or in a `[vision]` extras set.

## Related Issues

Closes #232 

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,10 +14,9 @@ dependencies = [
     "torch-spyre",
     "vllm",
 
-    # Listing as direct deps to ensure they're picked up from the correct index.
-    # Versions intentionally left out because these are dependencies of dependencies.
+    # Listing as a direct dep to ensure this is picked up from the pytorch-cpu index
+    # Version intentionally left out: Compatibility is determined by torch-spyre and vllm
     "torch",
-    "torchvision",
 ]
 requires-python = ">=3.11"
 dynamic = ["version"]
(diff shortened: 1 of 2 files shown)


ashokponkumar:
@kmehant PTAL

HarikrishnanBalagopal (approved):
LGTM

kmehant (approved):

PR #240, opened 2026-06-05
RFC: Port the upstream KV Connector experience to spyre-inference

wangchen615:
## Description

This is the first design doc for **#76 — [Epic] Develop KVCacheConnector for Spyre**.

It proposes how to wire `spyre-inference` into vLLM's upstream `OffloadingConnector` + multi-tier framework (`TieringOffloadingSpec` / `SecondaryTierManager` / `tiering/{fs,obj}`) so every KV-cache offload path that already targets `OffloadingConnector` works on Spyre with no per-connector plugin code.

The plan in two milestones:

- **M1** — register a `SpyreOffloadingSpec` via the upstream factory, reuse `CPUOffloadingManager` verbatim, and add one new `OffloadingHandler` pair (`SpyreCpuOffloadingHandlers`) backed by a single device↔host primitive `SpyreKvDmaCopier`. The copier is a thin wrapper around `torch_spyre._C.copy_tensor(src, dst, non_blocking=False)` — the public, stream-backed Spyre↔CPU copy entrypoint already bound in the dev-image-pinned torch-spyre commit (routes through `SpyreStream::copyAsync` → flex runtime DMA). Two new files of Spyre-specific code, no changes to `TorchSpyreWorker` or `TorchSpyrePlatform`. Single-tier host-RAM offload.
- **M1.5** — register a `SpyreTieringOffloadingSpec` (extends `SpyreOffloadingSpec`) so the upstream `tiering/fs`, `tiering/obj`, and any future `SecondaryTierManager` work on Spyre. ~50 LOC of glue.

Section 6 of the RFC has the full component design (§6.4 covers the M1.5 tiering spec); section 7 has the file-by-file plan; section 11 documents M2 (public `SpyreKvDmaCopier` import surface) and direct-device↔storage as out-of-scope items raised by review.

The RFC has gone through several substantive revisions during review:

- **2026-06-05** — initial draft.
- **2026-06-08** — rebased on the multi-tier framework that landed in vLLM v0.22 (PRs vllm-project/vllm#40020, #41735, #41968); incorporated review feedback from @yuezhu1.
- **2026-06-09 → 2026-06-11** — narrowed to a single M1.5 path after @yuezhu1 confirmed llmd_fs_backend maintainers are retiring the `MultiConnector + llmd_fs_backend` deployment shape; collapsed §4 around the public `_C.copy_tensor` primitive after @joerunde and @JRosenkranz pushed back on the dual-backend design and a closer read of `torch_spyre/csrc/spyre_stream.cpp` confirmed `copy_tensor` is what we want.

This is **docs-only** — no implementation in this PR. Marking ready for review now that the design has settled.

## Related Issues

- Tracks: #76 — [Epic] Develop KVCacheConnector for Spyre (this RFC is the first design doc under that epic).
- The RFC's section 11 enumerates the follow-up work that will become its own RFCs / issues (M2 public copier API, PD disaggregation, async DMA, stable on-device KV descriptor, direct device↔storage).

## Test Plan

Docs-only PR — no code paths exercised. The RFC's section 12 lists acceptance criteria for the eventual implementation:

**M1**

- `vllm serve` on Spyre with `OffloadingConnector` + `SpyreOffloadingSpec` runs a two-prompt sweep (second prompt extends the first by ≥256 tokens) and reports a non-zero `kv_offload_blocks_loaded` increment in worker metrics or the equivalent `OffloadingConnectorWorker: loading N blocks from host` log line.
- With `temperature=0`, generated tokens for both prompts are byte-identical to a baseline run with no `--kv-transfer-config`.
- `pytest tests/v1/kv_offload/` (three new files described in section 7) green on a Spyre runner.
- No source changes required to `TorchSpyreWorker` or `TorchSpyrePlatform`.

**M1.5**

- `SpyreTieringOffloadingSpec` + `tiering/fs` round-trips a known prompt-prefix across two `vllm serve` instances on the same node sharing a `root_dir`, and the second instance reports a prefix-cache hit (with `PYTHONHASHSEED` pinned identically on both — upstream `FileSystemTierManager` requirement).
- `tests/v1/kv_offload/test_tiering_spec.py` green on a CPU-only test runner using the upstream `example` secondary tier.
- M1.5 plugin-side LOC budget held: ≤50 LOC of glue on top of M1, excluding tests.

## Checklist

- [x] My code follows the project's code style (`uvx prek run --files docs/architecture/rfcs/upstream-connector-port.md` — all applicable hooks pass: `typos`, `markdownlint`, `ty check`. Ruff and actionlint correctly skipped — markdown-only change.)
- [x] I have updated the documentation (this PR _is_ the documentation)


docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,324 @@
...
+
+This RFC proposes how to combine the two: take the prototype's data-copy primitive, wrap it as an upstream-conformant `OffloadingHandler`, and register a `SpyreOffloadingSpec` so that every connector that already targets `OffloadingConnector` also works on Spyre.
+
+## 2. Goals and non-goals
+
+### Goals (M1)
+
+- A user runs vLLM on Spyre with `--kv-transfer-config '{"kv_connector":"OffloadingConnector", "kv_connector_extra_config":{"spec_name":"SpyreOffloadingSpec","cpu_bytes_to_use":"8000000000"}}'` and gets host-RAM offload that survives across requests.
+- All upstream connectors that compose `OffloadingConnector` (`SharedStorageOffloadingSpec`, llm-d-kv-cache `llmd_fs_backend`, etc.) work on Spyre with **no plugin-side handler changes** beyond what M1 ships.
```
yuezhu1: `SharedStorageOffloadingSpec` is part of `llmd_fs_backend`, if i remember correctly. Do we want to support any other connector besides `llmd_fs_backend` connector in M1, and later on?

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,324 @@
...
+- The Spyre device↔host copy goes through one named, testable primitive (`SpyreKvDmaCopier`) that wraps the prototype's `DmaiQPush`/`DmaoQPush` path (or its successor `torch_spyre._C.spyre_from_blob` path).
```
yuezhu1: since Spyre has its own D↔host copy calls, do we want to offer a functionality to allow other connectors to easily transplant their GPU based D↔host copy to Spyre? This would better help the usability. maybe a good target for M2

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,324 @@
...
+- `pytest tests/v1/kv_offload/` runs the same matrix as upstream for the CPU spec, plus a Spyre-specific test that round-trips a known-pattern block device→host→device.
```
yuezhu1, line 35: [tests/v1/kv_offload](https://github.com/vllm-project/vllm/tree/main/tests/v1/kv_offload) also includes test for tiering, which includes the test for fs and object tiering. Maybe we may need to clarify the supported data transfer path, then be more specific on what types of tiering tests that we want to include

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Non-goals (M1)
+
+- PD disaggregation. The prototype's NIXL adapter is reusable but lives in a follow-up RFC because it requires worker-to-worker NIXL agent setup that is orthogonal to the connector port.
+- Multi-tier (host RAM + file/object). Once the handler exists, layering `SharedStorageOffloadingSpec` on top is a configuration change, not code.
```
yuezhu1: 1. we may need to clarify what the supported data transfer paths are in M1? like device<->host RAM<->file/object, device<->host RAM, and device<->file/objec.
2. IMHO, we may not need big code changes in `llmd_fs_backend` for device<->host RAM to enable device<->host RAM in `llmd_fs_backend` handler. however, we need to carefully consider how to save/load data when having multi-tiers. how to make the decision on prefetching, async offloading, etc. Right now, i think data is saved to both connectors [ref1](https://github.com/vllm-project/vllm/blob/4200f621472925f8039191b9ff3b72951134cdf5/vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py#L304), and is loaded from the first connector that finds the match [ref2](https://github.com/vllm-project/vllm/blob/4200f621472925f8039191b9ff3b72951134cdf5/vllm/distributed/kv_transfer/kv_connector/v1/multi_connector.py#L395)

3. maybe not for current status, if we would like to allow the direct data transfer from device<->file/object in the future, we would need something similar to GDS, this would require changes in the `llmd_fs_backend`, as it mostly target for GPU.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,324 @@
...
+### 3.3 `CpuGpuOffloadingHandlers` (`vllm/v1/kv_offload/cpu/gpu_worker.py:375`)
+
+The reference CUDA implementation. It is **not directly reusable on Spyre** because:
+
+- It allocates `torch.cuda.Stream` per transfer (`spyre` has no public stream API; `_sync_device` in `TorchSpyreModelRunner` is a stub).
+- It asserts `gpu_tensor.is_cuda` on every registered KV tensor.
+- It calls `ops.swap_blocks_batch`, a custom CUDA op.
+- Optional `cudaHostRegister` pinning via `cudart()`.
+
+The factory layer above it (`OffloadingSpecFactory.register_spec`, `vllm/v1/kv_offload/factory.py:21`) **is** trivially reusable — it just imports a string-named class.
```
yuezhu1: maybe as 3.4 for dynamic module loading?

tdoublep:
Related issue #76

docs/architecture/rfcs/upstream-connector-port.md
yuezhu1: new multi-tiering pr for kv offloading in vllm through `llmd-fs-connector`. https://github.com/vllm-project/vllm/pull/44415

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Thanks — this drove the rebase pushed in 0ab42b5.

The short answer: M1 doesn't bind us to any single secondary connector, and `llmd_fs_backend` / `SharedStorageOffloadingSpec` aren't on the roadmap at all anymore. As of vLLM v0.22 the standalone `llmd-fs-connector` is end-of-life — its logic is now upstream as `vllm/v1/kv_offload/tiering/fs` (PR vllm-project/vllm#41735). The multi-tier framework that hosts it (PR vllm-project/vllm#40020) ships a generic `SecondaryTierManager` ABC + `SecondaryTierFactory`, and `tiering/obj` (PR vllm-project/vllm#41968) is already a second concrete tier in-tree.

So the new shape is:
- **M1**: `SpyreOffloadingSpec` covers device↔host RAM only — the primary tier. No secondaries. This is the *only* Spyre-specific piece.
- **M1.5**: `SpyreTieringOffloadingSpec` (extends M1) + secondaries selected by config. Once it lands, `fs`, `obj`, the upstream `example` tier, and anything anyone registers later via `SecondaryTierFactory` works on Spyre with zero per-tier plugin code.

I rewrote §1, §2, and §8 around this. §4a (new) has the data-paths-per-milestone table, and §6.3a (new) has the M1.5 spec design.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Yes — agreed, and I added this as **M2** in §11.

Concretely: M1 builds the primitive (`spyre_inference.v1.kv_offload.copier.SpyreKvDmaCopier`) but uses it only internally; M2 commits to its API surface so any third-party connector that today calls CUDA's `swap_blocks_batch` / `cudaMemcpy` can swap the device↔host hop for Spyre by importing one symbol. Adding it as an explicit milestone with you tagged as the proposer.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: All three points addressed in the rebase — thanks, these were the load-bearing pieces of feedback.

**(1) Supported transfer paths.** Added §4a with an explicit table:
- `Spyre device ↔ host RAM` — **M1**
- `Spyre device ↔ host RAM ↔ filesystem (tiering/fs)` — **M1.5**
- `Spyre device ↔ host RAM ↔ object store (tiering/obj)` — **M1.5**
- `Direct Spyre device ↔ filesystem / object store` — **out of scope** (see (3) below)

The first three all route through host RAM as the gateway, which is exactly what the upstream `SecondaryTierManager` contract assumes (see `tiering/base.py` — secondaries get a `primary_kv_view: memoryview` over CPU memory and cannot directly access device memory).

**(2) Multi-connector save-to-both / load-from-first.** Captured in §4a as a parenthetical FYI for users composing connectors (`MultiConnector` semantics — your `multi_connector.py:304` and `:395` refs). Important to note though that this isn't the cascade we're building: M1.5's tiered cache is internal to *one* `OffloadingConnector` and uses `TieringOffloadingManager`'s cascade-on-store / promote-on-load semantics, not `MultiConnector`'s save-to-both. The `MultiConnector` semantics only kick in if a user composes our `OffloadingConnector` with an unrelated connector (e.g., a NIXL PD connector) at engine init. Both are real, they just operate at different layers.

**(3) GDS-style direct device↔storage as future work.** Filed in §11 as a standalone out-of-scope item with explicit rationale: it would need both a torch-spyre primitive (Spyre analogue of GDS) *and* a contract change to upstream's `SecondaryTierManager` (drop the "primary_kv_view is over CPU memory" assumption). Two-sided change, properly scoped as a separate design.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Done — added as §3.4 "Dynamic spec loading" in the rebased version.

It explains `OffloadingSpecFactory.register_spec(name, module_path, class_name)` is lazy: registration only records a tuple, the module import happens in `create_spec(...)` when the user's `spec_name` selects this spec. That's the mechanism by which we can register `SpyreOffloadingSpec` from `spyre_inference/__init__.py` without dragging Spyre-only code into vLLM at import time, and CUDA-only deployments that load spyre-inference for unrelated reasons pay zero cost for our spec.

Same pattern documented for `SecondaryTierFactory.register_tier(...)` so M2 third-party tiers can register from their own packages.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Good catch on the acceptance criteria — rewrote them in 0469d28.

Two important corrections in this revision:

**1. The acceptance config is now your literal deployment config.** §12 now contains `vllm serve --kv-transfer-config '...'` invocations verbatim — A1.5.1 is the `MultiConnector` config you pasted, with `SpyreOffloadingSpec` slotted in as the first child connector and `SharedStorageOffloadingSpec` from `llmd_fs_backend.spec` as the second. Per-acceptance observable checks (server boots, files appear, restart-replay shows hits, cross-instance share works on the same mount).

**2. I had `llmd_fs_backend` framed wrong.** Looking at it again: only the *standalone* `llmd-fs-connector` package is EOL'd; the in-tree `llmd_fs_backend` inside `llm-d-kv-cache` (where `SharedStorageOffloadingSpec` lives) is alive and pinned to vLLM v0.22.x. So the deployment shape and the `tiering/fs` shape coexist during the transition window. §4a now reflects this as **Path A** (`MultiConnector`, today's shape, M1.5 primary acceptance) and **Path B** (`SpyreTieringOffloadingSpec` + `tiering/fs`, forward-looking).

**3. Load-bearing question I added (§10 Q7):** Reading `llmd_fs_backend/spec.py` and `worker.py`, `SharedStorageOffloadingSpec` extends `OffloadingSpec` directly (no `is_cuda_alike` gate, unlike `CPUOffloadingSpec`), and `StorageOffloadingHandlers` accepts a generic `GPULoadStoreSpec` tag. Combined with the fact that `MultiConnector.register_kv_caches` hands raw `kv_caches` to every child, the second connector receives Spyre device tensors. *If* `StorageOffloadingHandlers`' device→host hop is owned by the first connector (us) — which it appears to be from the worker code — Path A works on Spyre with **zero** changes to `llmd_fs_backend`. Verifying that is A1.5.1's first checkbox; if it doesn't, the fix is a small upstream PR to `llmd_fs_backend` rather than Spyre-side glue.

I left the multi-connector save-to-both / load-from-first FYI in §4a as the explanation of the `MultiConnector` cascade semantics (your `multi_connector.py:304`/`:395` refs).

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Quick correction to my earlier reply on this thread — `SharedStorageOffloadingSpec` and `llmd_fs_backend` aren't actually deprecated, only the *standalone* `llmd-fs-connector` Python package is. The in-tree `llmd_fs_backend` in `llm-d-kv-cache` is alive and pinned to vLLM v0.22.x, and §4a / §12 now reflect it as **Path A** (the `MultiConnector` shape that llm-d v0.8 deployments use today). `tiering/fs` is **Path B** (the upstream-canonical shape post-v0.22). Both M1.5-in-scope; A1.5.1 acceptance is the verbatim deployment config.

Apologies for the earlier framing — the rebase commit was right that the framework moved upstream, but I overstated the EOL.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -108,14 +108,18 @@ The prototype's NIXL connector and its `CpuBufferManager` are about getting two
...
+| Spyre device ↔ host RAM ↔ filesystem/object (Path B) | M1.5+ | `OffloadingConnector` + `SpyreTieringOffloadingSpec` + `tiering/fs` or `tiering/obj` | The v0.22+ canonical shape. Internal cascade (`TieringOffloadingManager`), not save-to-both. Supersedes the standalone `llmd-fs-connector` (final standalone release `v0.22`). |
+| Direct Spyre device ↔ filesystem / object store | Out of scope | n/a | Would require a Spyre-side analogue of NVIDIA GDS so a secondary tier can DMA without a host bounce. Not provided by torch-spyre today, and the upstream `SecondaryTierManager` contract assumes the `primary_kv_view` is over CPU memory; supporting this would change both. Filed as a future-work item. |
+
+**Why Path A and Path B are both M1.5 in scope.** Path A is what's deployed today across llm-d v0.8 clusters. Path B is the upstream-canonical direction. Yue's example config (in the M1.5 acceptance criterion below) is Path A verbatim. Both paths reuse the same `SpyreOffloadingSpec` from M1 — the only difference is what sits on the other side of host RAM, which is configured per-deployment, not coded in the plugin.
```
yuezhu1: I would recommend to highlight the difference between Path A and B. add my understanding for the difference between Path A and Path B below. 

**Path A (MultiConnector)** uses two independent OffloadingConnectors that operate in parallel without coordination: saves fan out to both CPU and storage simultaneously, and loads return from whichever connector reports a hit first. 

**Path B (TieringOffloadingSpec)** uses a single unified system where the TieringOffloadingManager orchestrates hierarchical data flow: GPU transfers always go through the CPU primary tier, which then cascades to secondary storage tiers, ensuring coordinated state management and staged promotion.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,527 @@
...
+
+| File | Coverage |
+|---|---|
+| `test_copier_round_trip.py` | Allocate a Spyre tensor with a known fp16 pattern, copy d2h, mutate host copy, copy h2d, assert content. Skipped if `device("spyre")` not available (CI gating already exists for other Spyre tests). |
+| `test_spec_registration.py` | Import `spyre_inference`, then `OffloadingSpecFactory.create_spec(...)` resolves. Pure-CPU test — no Spyre device required. |
+| `test_handler_dispatch.py` | Exercise `device_to_host_handler` / `host_to_device_handler` against `(src, dst)` tuples and assert the correct content lands. |
+
+### M1.5 files (incremental)
+
+| File | Purpose | Approx LOC |
```
yuezhu1, line 957: do we want to include the test for  `SpyreOffloadingSpec` (D↔H) with `SharedStorageOffloadingSpec` from `llmd_fs_backend` (H↔FS) ?

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,527 @@
...
+
+After M1 + M1.5 ship, the following work on Spyre **without further plugin code**:
+
+- **Single-tier host-RAM offload** (M1) — via `SpyreOffloadingSpec`. Same prefix-cache semantics as the upstream CPU spec on CUDA.
+- **`tiering/fs` secondary tier** (`vllm/v1/kv_offload/tiering/fs/manager.py:FileSystemTierManager`) — pure-Python disk-backed tier with content-hashed paths via `FileMapper`. With matching `PYTHONHASHSEED`, two Spyre instances on a shared `root_dir` (e.g. RWX PVC) cross-share blocks. Replaces the standalone `llmd-fs-connector` (EOL as of v0.22).
+- **`tiering/obj` secondary tier** (PR #41968) — object-store backend (S3-style).
+- **`tiering/example` secondary tier** — in-memory tier shipped upstream as a reference implementation; we use it in tests.
+- **Future secondary tiers** — anything someone registers with `SecondaryTierFactory.register_tier(name, module, class)` works on Spyre as soon as it works on CUDA, since the Spyre-specific code stops at the primary tier.
+- **LMCache offloading** — talks to the same `OffloadingHandler` API at the device↔host seam; M1 alone is enough.
```
yuezhu1: maybe we need to be more specific here for LMCache offloading? they offer lots of different connectors. we may want to mention that the proposed design could work for connectors supporting device↔host transfer via `OffloadingHandler`

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,527 @@
...
+    "cpu_bytes_to_use": 8000000000,
+    "lazy_offload": true
+  }
+}'
+```
+
+- [ ] Server boots. `OffloadingConnectorWorker.register_kv_caches` is reached on the Spyre worker without raising.
+- [ ] A 1k-prompt sweep with overlapping prefixes reports a non-zero `kv_offload_blocks_evicted` (or the equivalent prefix-cache-hit metric exposed by `OffloadingConnector` in this vLLM version).
+- [ ] Generated outputs match a non-offloading baseline run on the same prompts (allow ≤1e-3 token-distribution drift; with `temperature=0` outputs should be identical).
+
```
yuezhu1, line 1042: do we want to include loading test too? like complete run with store/load operations

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Adopted in d12ffaf — §4a now leads with your Path A (independent, save-fan-out, load-from-first) vs. Path B (single TieringOffloadingManager orchestrating a hierarchy) distinction, with attribution to you on the formulation. The phrasing "two write-through caches with MultiConnector deciding lookups" vs. "a real tiered cache with one manager" is what I went with operationally. Also added an explicit note that SharedOffloadRegion is a Path-B-only concept (not load-bearing for the deployed Path A shape). Let me know if any of the framing still misses.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Good catch — added in d12ffaf. The §7 M1.5 file table now includes `test_multiconnector_register.py`, a cheap Spyre-runner pre-check that constructs `MultiConnector(SpyreOffloadingSpec, SharedStorageOffloadingSpec)` with `llmd_fs_backend` installed and asserts both children's `register_kv_caches` calls succeed against a real Spyre kv_caches dict. ~80 LOC, runs in seconds, no model warmup needed. This is the cheap falsifier for §10 Q7: if it fails, A1.5.1 will fail, so we want to know early. A1.5.3 now requires both this and `test_tiering_spec.py` green.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Tightened in d12ffaf — §8's LMCache bullet now reads "LMCache connectors that route through the OffloadingHandler device↔host seam," and explicitly notes that LMCache ships several connector flavors, not all of which use this seam (some have their own CUDA copy path). M1 supports the ones that do; the others would need an LMCache-side change to swap their device↔host hop for SpyreKvDmaCopier — which is the M2 use case (§11). Thanks, this needed precision.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Yes — added in d12ffaf. A1.5.1 now has explicit load-side checks, broken into:

- **Same-instance load:** restart with same config + same PYTHONHASHSEED, send a second prompt sharing ≥256 tokens with the warmup, observe `kv_offload_blocks_loaded` increment attributable to the second connector (or the equivalent counter from `llmd_fs_backend/metrics.py`), tokens identical at temperature=0.
- **Cross-instance load:** second host mounting the same RWX volume with the same PYTHONHASHSEED, observe a storage-tier hit on the *first* request (no warmup needed on the second host).
- **PYTHONHASHSEED negative case:** documented as a one-time check (different seeds → different filenames → must miss), recorded in the impl issue rather than as a recurring CI gate.

Also tightened A1.1 the same way — replaced the vague "1e-3 token-distribution drift" with byte-identical token equality at temperature=0, and named the metric/log-line so the test harness can pick one. PTAL.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Thanks — addressed across the rebase (0ab42b5) and follow-ups:

**Supported transfer paths** are now enumerated explicitly in §4a (added in 0ab42b5, sharpened in d12ffaf):
- Spyre device ↔ host RAM (single tier) — **M1**
- Spyre device ↔ host RAM ↔ filesystem (Path A: `MultiConnector` + `llmd_fs_backend`) — **M1.5**
- Spyre device ↔ host RAM ↔ filesystem/object (Path B: `SpyreTieringOffloadingSpec` + `tiering/{fs,obj}`) — **M1.5**
- Direct device ↔ filesystem/object — **out of scope** (would need a Spyre GDS analog)

**Tiering tests we want to include** (§7 M1.5 file table):
- `test_tiering_spec.py` (~120 LOC, CPU-only) — pure-CPU test that builds `SpyreTieringOffloadingSpec` with the upstream `example` secondary tier and asserts store/load through the cascade matches upstream promotion semantics. This is the Path B unit-level coverage.
- `test_multiconnector_register.py` (~80 LOC, Spyre-runner) — `MultiConnector(SpyreOffloadingSpec, SharedStorageOffloadingSpec)` register-only check; the cheap falsifier for §10 Q7 before A1.5.1's full integration. Path A pre-check.
- A1.5.1 / A1.5.2 in §12 are the end-to-end `vllm serve` integration coverage for Path A and Path B respectively, with explicit store-side and load-side (same-instance + cross-instance) assertions.

We rely on upstream's own `tests/v1/kv_offload` to cover the *internals* of `tiering/{fs,obj}` and `TieringOffloadingManager` — our plugin-side tests cover only the Spyre-specific seams (device↔host transfer, spec registration, `MultiConnector` composition with our spec).

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -16,26 +17,33 @@ The upstream vLLM `OffloadingConnector` framework gives every CUDA platform thre
...
 
-This RFC proposes how to combine the two: take the prototype's data-copy primitive, wrap it as an upstream-conformant `OffloadingHandler`, and register a `SpyreOffloadingSpec` so that every connector that already targets `OffloadingConnector` also works on Spyre.
+This RFC proposes how to combine the two: take the prototype's data-copy primitive, wrap it as an upstream-conformant `OffloadingHandler`, and register a `SpyreOffloadingSpec` so that the upstream `OffloadingConnector` works on Spyre. A small follow-on (`SpyreTieringOffloadingSpec`) extends the same handler with the `SecondaryTierManager` hook, so every secondary tier registered with `SecondaryTierFactory` (`fs`, `obj`, anything added later) works on Spyre with no per-tier plugin code.
```
yuezhu1: do we want to brief touch the reason for no secondary tier plugin code change on Sypre? We could simply mention that like 

"every secondary tier registered with `SecondaryTierFactory` (`fs`, `obj`, anything added later) works on Spyre with no per-tier plugin code, as they only interact with Primary(CPU)↔Secondary(Storage) transfers."

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -16,26 +17,33 @@ The upstream vLLM `OffloadingConnector` framework gives every CUDA platform thre
...
-### Non-goals (M1)
+### Goals (M1.5 — small follow-on, scoped here)
+
+- A user runs vLLM on Spyre with `spec_name: "SpyreTieringOffloadingSpec"` plus a `secondary_tiers: [{type: "fs", root_dir: "/mnt/kvcache"}]` entry and gets host-RAM-plus-FS tiered offload, with content-hashed paths that two instances on a shared volume cross-share.
+- Every `SecondaryTierManager` registered with `SecondaryTierFactory` (`fs`, `obj`, future tiers) works on Spyre with **no per-tier plugin code** beyond what M1.5 ships.
+- M1.5 lands at most ~50 LOC of glue on top of M1; if it grows, the design is wrong — pause and revise.
+
+### Non-goals (M1 + M1.5)
 
 - PD disaggregation. The prototype's NIXL adapter is reusable but lives in a follow-up RFC because it requires worker-to-worker NIXL agent setup that is orthogonal to the connector port.
```
yuezhu1: maybe consider remove this section, since it is mentioned in out of scope section already? or can also merge some into the out of scope section.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -92,7 +104,18 @@ host = buf.to('cpu')        # device → host
 
 ### 4.3 What's NIXL-specific (out of scope for M1)
```
yuezhu1: i would recommend to remove section4.3

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+host = buf.to('cpu')        # device → host
+```
+
+`get_dmpa` returns the device physical address of an existing Spyre tensor; `spyre_from_blob` constructs a Spyre tensor view at an arbitrary device address. The prototype's accessor uses these to read/write at flit-offset granularity from Python without touching senlib. **This is the path we should prefer for the M1 handler** because it operates on `torch.Tensor` objects (so `OffloadingHandler` doesn't need to know about `DmaoDataCB`) and because it does not depend on a pre-built libsenlib being available in the deployment image.
+
+### 4.3 What's NIXL-specific (out of scope for M1)
+
+The prototype's NIXL connector and its `CpuBufferManager` are about getting two *hosts* to exchange those CPU tensors over the network. They are independent of how the Spyre-side data got into a CPU tensor in the first place. M1 reuses none of this.
+
+## 4a. Data paths in scope
```
yuezhu1: section 4.4?

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+| Spyre device ↔ host RAM ↔ filesystem/object (Path B) | M1.5+ | `OffloadingConnector` + `SpyreTieringOffloadingSpec` + `tiering/fs` or `tiering/obj` | The v0.22+ canonical shape. Internal cascade (`TieringOffloadingManager`), not save-to-both. Supersedes the standalone `llmd-fs-connector` (final standalone release `v0.22`). |
+| Direct Spyre device ↔ filesystem / object store | Out of scope | n/a | Would require a Spyre-side analogue of NVIDIA GDS so a secondary tier can DMA without a host bounce. Not provided by torch-spyre today, and the upstream `SecondaryTierManager` contract assumes the `primary_kv_view` is over CPU memory; supporting this would change both. Filed as a future-work item. |
+
+**Path A vs. Path B — the architectural difference, not just the config difference** (formulation per [@yuezhu1](https://github.com/yuezhu1) on the M1.5 draft):
+
```
yuezhu1, line 276: are we targeting both Path A and B for M1.5? if `llmd_fs_backend` is stopped as `lmd-fs-connector==0.22` as final release of the standalone connector, do we want to support Path B only?

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+- **`tiering/obj` secondary tier** (PR #41968) — object-store backend (S3-style).
+- **`tiering/example` secondary tier** — in-memory tier shipped upstream as a reference implementation; we use it in tests.
+- **Future secondary tiers** — anything someone registers with `SecondaryTierFactory.register_tier(name, module, class)` works on Spyre as soon as it works on CUDA, since the Spyre-specific code stops at the primary tier.
+- **LMCache connectors that route through the `OffloadingHandler` device↔host seam** — M1 alone is enough. LMCache ships several connector flavors, not all of which use this seam (some implement their own CUDA copy path); M1 supports the ones that do, and the others would need an LMCache-side change to swap their device↔host hop for `SpyreKvDmaCopier` (this is the M2 use case in §11).
+
+The only connector that does **not** drop in is anything that requires async copy semantics (e.g. CUDA-graph-capturable transfers). None of the M1/M1.5-relevant tiers do — the upstream `SecondaryTierManager` contract is explicitly async-via-job-poll, not async-via-CUDA-events.
+
+## 9. Migration: from the prior PD prototype to upstream
```
yuezhu1, line 981: Since PD is out of scope for now, i would recommend to merge this into out of scope section for migration discussion.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+**A1.2 — plugin-side test suite green.**
+
+- [ ] `pytest spyre_inference/tests/v1/kv_offload/test_copier_round_trip.py` passes on a Spyre runner.
+- [ ] `pytest spyre_inference/tests/v1/kv_offload/test_spec_registration.py` and `test_handler_dispatch.py` pass on CPU-only runners.
+
+**A1.3 — no plugin-platform-side regressions.**
+
+- [ ] No source changes required to `TorchSpyreWorker` or `TorchSpyrePlatform` for M1 to land. (If we have to change them, the RFC's premise is wrong — pause and revise.) Verified by inspecting the M1 PR diff: `spyre_inference/v1/worker/` and `spyre_inference/platform.py` are unchanged.
+- [ ] The existing Spyre platform/worker test suite (`pytest spyre_inference/tests/ -k 'not kv_offload'`) passes both with `SpyreOffloadingSpec` registered (M1 default after `spyre_inference` is imported) and with the connector unselected (no `--kv-transfer-config`). Same suite, two configs, both green — confirms registration alone has no effect when the connector isn't selected.
+- [ ] `bash format.sh` clean.
```
yuezhu1: what is this `bash format.sh` for

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+- [ ] **Load side, same instance.** Restart the server with the same config, same model, and the same `PYTHONHASHSEED` value as the warmup run. Send a *second* prompt that shares its first ≥256 tokens with the warmup prompt. Worker log emits a `kv_offload_blocks_loaded` increment attributable to the *second* connector (or its `llmd_fs_backend`-side equivalent — `llmd_fs_backend` exposes its own `metrics.py` counters). Generated tokens for the new prompt are byte-identical to a no-cache baseline at `temperature=0`.
+- [ ] **Load side, cross-instance.** On a second host that mounts the same `/mnt/files-storage` (RWX volume), start a fresh `vllm serve` with the same config and the *same* `PYTHONHASHSEED` value. Issue the same prefix-sharing prompt. Observe a storage-tier hit on the *first* request (no warmup needed on the second host). Tokens identical.
+- [ ] **`PYTHONHASHSEED` failure mode is documented.** If `PYTHONHASHSEED` differs across instances, the cross-instance step above MUST miss (different hash chain → different filenames). Run this once with mismatched seeds to confirm the negative case — record the result in the M1.5 implementation issue, not as a recurring CI gate.
```
yuezhu1: feel this may not be needed. if `PYTHONHASHSEED` differs among instances, other instance cannot see it. This is from vllm not from this RFC.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,558 @@
...
+}'
+```
+
+- [ ] Server boots; the warmup-restart-replay test from A1.5.1 reports a prefix-cache hit on the second run (with `PYTHONHASHSEED` pinned identically on both invocations).
+- [ ] Cross-instance share works on a shared `root_dir`.
+
+**A1.5.3 — plugin-side test suite green.**
+
+- [ ] `pytest spyre_inference/tests/v1/kv_offload/test_tiering_spec.py` passes on CPU-only runners using the upstream `example` secondary tier (Path B coverage).
+- [ ] `pytest spyre_inference/tests/v1/kv_offload/test_multiconnector_register.py` passes on a Spyre runner with `llmd_fs_backend` installed (Path A pre-check, see §10 Q7). If it fails, A1.5.1 will fail — fixing this is the gate to running A1.5.1.
```
yuezhu1: If we want to exclude Path A, we would need to revisit this

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,596 @@
...
+class SpyreKvDmaCopier:
+    """Single-purpose owner of every host↔Spyre KV byte transfer.
+
+    Wraps two backends:
+      - 'spyre_from_blob'  : torch_spyre._C.{get_dmpa, spyre_from_blob}.
+                             Operates on torch.Tensor inputs; preferred.
+      - 'senlib_dma'       : libsenlib DmaiQPush/DmaoQPush over flit_offset
+                             addresses; matches the demonstrated PD-disagg
+                             plugin. Used when 'spyre_from_blob' isn't
+                             available (older runtime images).
```
joerunde: I'm probably not fully understanding this but this wording sounds like we're planning for backwards compatibility for older runtime stacks. Any reason to not just support `spyre_from_blob`?

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Confirmed and acted on in 43d2d58. Per your read on the Haifa team's direction, Path A is no longer in scope — the RFC now targets only the upstream-canonical `SpyreTieringOffloadingSpec` + `tiering/fs`/`tiering/obj` shape.

Reasoning, for the record: vllm-project/vllm#41735 already upstreamed the FS storage code (`tiering/fs`); vllm-project/vllm#44415 (merged 2026-06-09) added the end-user usage docs for the multi-tier shape. So new deployments will write `TieringOffloadingSpec` configs from day one, and the prior `MultiConnector + llmd_fs_backend` shape only exists for migration-from-llm-d-v0.8 readers. The RFC keeps a one-paragraph historical note in §3.5 for those readers but does not target the prior shape.

Cascading edits in 43d2d58:
- §4.3 (renumbered from §4a): single M1.5 row, Path A row gone.
- §6.4 (renumbered from §6.3a) and §10 Q6: dropped Path-B-only qualifiers (it's the only path now).
- §10 Q7 deleted entirely (only existed to gate Path A).
- §7: `test_multiconnector_register.py` removed (was the Q7 falsifier).
- §12: A1.5.1 is now the `SpyreTieringOffloadingSpec + tiering/fs` config end-to-end; the old Path A acceptance config is gone.
- §13: added vllm-project/vllm#44415 reference; kept llm-d-kv-cache pointer but as historical context only.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Done in 43d2d58 — Path A is now excluded. The §13 `llm-d-kv-cache` reference stayed but is reframed as historical context for readers migrating from existing llm-d v0.8 deployments. Also added vllm-project/vllm#44415 to §13 as the canonical upstream user-facing guide for the multi-tier shape M1.5 targets.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Agreed — removed in 43d2d58. `PYTHONHASHSEED` mismatch breaking cross-instance share is upstream's contract (it's documented on `FileSystemTierManager` itself), not something this RFC's acceptance gates should re-prove. The remaining `PYTHONHASHSEED` mentions in A1.5.1 are now just preconditions for the same-instance and cross-instance load checks, no negative-case documentation step.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Clarified inline in 43d2d58: `format.sh` at the repo root is this repo's lint wrapper around `uvx prek` (it runs `--all-files` if no args are given). The acceptance line now reads:  `bash format.sh` clean. (`format.sh` at the repo root is this repo's lint wrapper around `uvx prek`; runs `--all-files` if no arg is given.)

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Done in 43d2d58 — the §9 PD paragraph is gone; its content is now a sub-bullet under "PD disaggregation on Spyre" in §11 ("Every component PD needs *except* the cross-host transport is delivered by M1...").

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Renumbered in 43d2d58 — "§4a" → "§4.3". Worked out cleanly because the same commit dropped the prior §4.3 (NIXL out-of-scope, your other comment 3389029011) so the slot freed up. The §6 subsections renumbered too (6.3a → 6.4, 6.4 → 6.5, 6.5 → 6.6) so the spec / registration / worker-glue sequence stays consecutive.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Removed in 43d2d58. The NIXL out-of-scope content was already covered under "PD disaggregation on Spyre" in §11, and now the prior-prototype's NIXL-vs-CpuBufferManager detail (about getting two *hosts* to exchange CPU tensors) is folded into that §11 bullet too — single source of truth for PD-related out-of-scope items.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Done in 43d2d58 — §2 non-goals subsection removed entirely. The PD-disagg, flit-offset, and (no-longer-relevant-anyway) Path A items were all duplicated in §11; §2 now ends with one line: *"Items explicitly out of scope (PD disaggregation, replacing the flit-offset addressing scheme, etc.) are listed in §11 alongside their owners and follow-up plans."*

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Adopted close to verbatim in 43d2d58. The §1 motivation paragraph now ends:  "...so every secondary tier registered with `SecondaryTierFactory` (`fs`, `obj`, anything added later) works on Spyre with no per-tier plugin code, since secondary tiers only interact with primary(CPU)↔secondary(storage) transfers — they never touch device tensors. The Spyre-specific code stops at the device↔host primary tier, and everything above it is platform-agnostic upstream code."  Thanks — this answers the *why* without making the reader trace through §3.5 / §4 first.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Update: in 43d2d58 `test_multiconnector_register.py` was removed along with the rest of Path A. With the `MultiConnector + llmd_fs_backend` shape no longer in scope (per 3389104355), the test had no remaining purpose — it was specifically the cheap falsifier for §10 Q7, which is also gone. M1.5 testing is now `test_tiering_spec.py` only (§7), with end-to-end coverage via A1.5.1's `SpyreTieringOffloadingSpec + tiering/fs` integration.

docs/architecture/rfcs/upstream-connector-port.md
```diff
@@ -0,0 +1,559 @@
...
+
+```python
+self._C = importlib.import_module("torch_spyre._C")
+addr = self._C.get_dmpa(x_spyre)
+buf  = self._C.spyre_from_blob(addr, size=..., stride=..., dtype=torch.float16)
+buf.copy_(data)             # host → device
+host = buf.to('cpu')        # device → host
+```
+
+`get_dmpa` returns the device physical address of an existing Spyre tensor; `spyre_from_blob` constructs a Spyre tensor view at an arbitrary device address. The prototype's accessor uses these to read/write at flit-offset granularity from Python without touching senlib. **This is the path we should prefer for the M1 handler** because it operates on `torch.Tensor` objects (so `OffloadingHandler` doesn't need to know about `DmaoDataCB`) and because it does not depend on a pre-built libsenlib being available in the deployment image.
```
JRosenkranz: would this only work in PF mode?

JRosenkranz: For spyre_from_blob, does this still go through SpyreAllocator?

wangchen615: Good question — and the answer that drove a wider RFC pivot in [`e85097b`](https://github.com/torch-spyre/spyre-inference/pull/240/commits/e85097b).

Short answer: **the M1 design no longer uses `spyre_from_blob` or any DMPA-address path** — so the PF-mode question becomes moot for this RFC. Instead, M1's `SpyreKvDmaCopier` is a thin wrapper around the public `torch_spyre._C.copy_tensor(src, dst, non_blocking=False)`, which is bound to [`SpyreStream::copyAsync`](https://github.com/torch-spyre/torch-spyre/blob/4dcfee15c3a9344652f067149ec65c4bf2941890/torch_spyre/csrc/spyre_stream.cpp#L142) and routes through the flex runtime — no PF / senlib direct addressing involved.

What changed: as you suggested, I re-read `spyre_stream.cpp` and `spyre_mem.cpp` at the dev-image-pinned commit `4dcfee15`, and confirmed `copy_tensor` is the right primitive. `§6.1` now wraps `copy_tensor` directly; `§4` calls out the prototype's senlib + DMPA-style paths as predecessors no longer used.

wangchen615: Same answer: this question is no longer load-bearing for the RFC — M1 doesn't use `spyre_from_blob`. As of [`e85097b`](https://github.com/torch-spyre/spyre-inference/pull/240/commits/e85097b), `SpyreKvDmaCopier` wraps `torch_spyre._C.copy_tensor(src, dst, non_blocking=False)` instead.

Both arguments to `copy_tensor` are existing `at::Tensor` objects whose Spyre-side storage is already SpyreAllocator-managed (the connector's host-side block tensors come from `torch.empty(...)`; the Spyre-side KV cache tensors are what `_allocate_kv_cache_tensors` already returned via the standard allocator path). The C++ side just resolves the `SharedOwnerCtx` / `composite_addr` from the Spyre tensor's storage and DMAs to/from the CPU buffer — see [`spyre_mem.cpp:581`](https://github.com/torch-spyre/torch-spyre/blob/4dcfee15c3a9344652f067149ec65c4bf2941890/torch_spyre/csrc/spyre_mem.cpp#L581) for `spyre_copy_from` and [`spyre_stream.cpp:142`](https://github.com/torch-spyre/torch-spyre/blob/4dcfee15c3a9344652f067149ec65c4bf2941890/torch_spyre/csrc/spyre_stream.cpp#L142) for `copyAsync`.

So nothing in the M1 flow constructs a tensor view at an arbitrary device address — every Spyre-side tensor we touch came from SpyreAllocator already. PTAL.

docs/architecture/rfcs/upstream-connector-port.md
wangchen615: Thanks for the comment.

@JRosenkranz also pointed out the runtime had been updated, and on a closer read of `torch_spyre/csrc/spyre_stream.cpp` at the dev-image-pinned commit (`4dcfee15`), the right answer to your original question turned out to be even simpler.

Pushed in [`e85097b`](https://github.com/torch-spyre/spyre-inference/pull/240/commits/e85097b): **`SpyreKvDmaCopier` is now a thin wrapper around a single public entrypoint, `torch_spyre._C.copy_tensor(src, dst, non_blocking=False)`**, which routes through [`SpyreStream::copyAsync`](https://github.com/torch-spyre/torch-spyre/blob/4dcfee15c3a9344652f067149ec65c4bf2941890/torch_spyre/csrc/spyre_stream.cpp#L142) → flex runtime DMA. No `senlib_dma` fallback, no `spyre_from_blob`, no `SPYRE_KV_DMA_BACKEND` env var, no flit-offset addressing.

What changed since the previous reply:
- Earlier I framed senlib_dma as "the only working backend in the current dev image" because I'd grepped for `spyre_from_blob`/`get_dma_address` and seen they weren't exported. That was the right grep but the wrong question — there's already a higher-level public entrypoint (`copy_tensor`) that does what we need, bound at [`module.cpp:272`](https://github.com/torch-spyre/torch-spyre/blob/4dcfee15c3a9344652f067149ec65c4bf2941890/torch_spyre/csrc/module.cpp#L272).
- §6.1 is now ~30 LOC instead of ~120; §4 collapses the two prototype-DMA subsections into a single `copy_tensor` description; §10 Q1 is marked Resolved; §11's "Upstream the DMPA accessors" follow-up is removed (no longer needed).

Net: your original instinct ("why two backends?") was right, and the answer is now "one backend, on the public copy primitive that's already in the dev image" — which is what you were pushing for. Thanks for it.

PR #249, opened 2026-06-09
Reworked CustomOp layers with data transfer wrapped

### LINKED ISSUE #244: [Feature]: Wrap data transfers for torch.compile
### Feature description

With this feature data transfers between `cpu` and `spyre` will be wrapped in an opaque layer for torch.compile.

### Motivation and context

Torch.compile cannot trace through data transfers between host and device, for example done through `.to(...)` or `convert(...)`. Unfortunately, such data transfers are currently common, because of unsupported slicing operations on `spyre`. Therefore, they need to be wrapped for `torch.compile` to avoid tracing. This mostly affects some of our `CustomOps`, as well as the `SpyreModelRunner`.

### Proposed solution

_No response_

### Checklist

bohnstingl:
## Description

In this PR, we rework how the data transfers are handled and exposed in the case when torch.compile will be used. In particular, the `convert` operation is masked behind an opaque layer for torch.compile and included as-is into the graph without tracing it. Wrapping the `convert` operation directly allows to simplify the CustomOp implementation that require data movement, as otherwise an individual wrapper for every CustomOp needs to be implemented.

## Related Issues

This PR aims to resolve https://github.com/torch-spyre/spyre-inference/issues/244

## Test Plan

Change transparent to the user and to the tests. No new tests are required. Tests might change once torch.compile is fully enabled.

## Checklist


bohnstingl:
bot:next-test

bohnstingl:
bot:next-test

bohnstingl:
bot:next-test

bohnstingl:
Unclear whether CI fails are related. Tests pass locally for me

joerunde (approved):
Lpgtm!

I merged in main and just added a tiny update to call `RotaryEmbedding.forward_static` with kwargs instead of positional args, in case future vllm updates re-order the arguments. (The apis there are not very stable)

joerunde:
Merging in main caused a regression, debug logs here:

# TP=2 generate-matches test failure after merging main into convert_wrapper

> **Status: ✅ resolved** &nbsp;&nbsp;·&nbsp;&nbsp; Created 2026-06-18 19:45:21

## Header

- **Cluster slug:** `tp2-matches-merge`
- **Symptom:** `NotImplementedError: Could not run 'aten::index_select' with arguments from the 'spyre' backend` raised inside our `spyre_rotary_cpu` custom op when `RotaryEmbedding.forward_static` indexes `cos_sin_cache`. Fires on the very first `_dummy_run` during engine warmup, on TP=1 and TP=2 alike.
- **Branch:** `convert_wrapper` @ `8eac9c9` (post merge of `main`)
- **Failing test node ids in cluster:**
  - ⭐ `tests/test_distributed_tp2.py::test_tp2_llm_generate_matches_tp1` *(probe)*
- **Repro:** `uv run pytest tests/test_distributed_tp2.py -m "not upstream" -k matches -svx`

## Session summary

**Root cause.** The merge of `main` bumped the pinned `torch-spyre` revision from `4dcfee15` to `a6baa9c7` (uv.lock change). The new revision installs a monkey-patch on `nn.Module.to` in `torch_spyre/model_utils.py:_spyre_module_to`. When `spyre_model_runner._load_model()` calls `self.model.to(device=self._spyre_device)` at line 267, the patched `to` reroutes into `load_model_to_spyre()`, which iterates `module._buffers` directly and assigns `module._buffers[buf_name] = _dma_to_spyre_default(buf, ...)`. That path **never calls `_apply`**, so `SpyreRotaryEmbedding._apply` returning `self` has no effect — `cos_sin_cache` gets DMA'd to Spyre anyway. Inside `_rotary_cpu_op_func`, `RotaryEmbedding.forward_static` then calls `cos_sin_cache.index_select(0, positions)` where `cos_sin_cache` is on Spyre and `aten::index_select` has no Spyre kernel → `NotImplementedError`.

**Fix.** In `SpyreRotaryEmbedding.__init__`, hold a CPU-side reference to the freshly computed `cos_sin_cache` as a regular attribute (`self._cpu_cos_sin_cache = self.cos_sin_cache`). The CPU op body now passes `layer._cpu_cos_sin_cache` to `forward_static` instead of `layer.cos_sin_cache`. Regular attributes are not in `_buffers`, so `load_model_to_spyre` never touches them. The dead `_apply` override was removed.

## Hypothesis / fix queue

| # | Hypothesis / experiment | Status | Outcome | Evidence |
|---|---|---|---|---|
| 1 | Print devices of `positions`, `cpu_positions`, `layer.cos_sin_cache` at top of `_rotary_cpu_op_func` | ✅ confirmed | `cos_sin_cache.device=spyre:0` at op-body entry. Confirms the failing tensor; rules out "positions still on Spyre" red herring. | `logs/tp2-matches-merge/debug-print.txt` |
| 2 | `_match_cos_sin_cache_dtype` (the only place in vLLM that reassigns the cache buffer) is the culprit. Wrap it in a tracer. | ❌ ruled out | `_match_cos_sin_cache_dtype` never called. Counter stays at zero. | `logs/tp2-matches-merge/trace-match.txt` |
| 3 | Our `SpyreRotaryEmbedding._apply` is being called but somehow not preventing the move. Tracer on it. | ❌ ruled out | `_apply` never called either. Cache moves `cpu` → `spyre` without going through any nn.Module hook our class sees. | `logs/tp2-matches-merge/trace-apply-stack.txt` |
| 4 | Buffer reassignment via `nn.Module.__setattr__`. Tracer on setattr for `cos_sin_cache`. | ❌ ruled out | No `__setattr__` hits either — confirms the move is via direct `_buffers[name] = ...` dict assignment. | `logs/tp2-matches-merge/trace-setattr.txt` |
| 5 | Subclass `_buffers` dict with a traced `__setitem__` to capture the stack at the moment of the move. | ✅ confirmed | Stack: `self.model.to(device=self._spyre_device)` → `torch_spyre/model_utils.py:254 _spyre_module_to` → `load_model_to_spyre` → `module._buffers[buf_name] = _dma_to_spyre_default(buf, ...)`. **Smoking gun.** | `logs/tp2-matches-merge/trace-bufset.txt` |
| 6 | Fix: hold a CPU reference in `__init__` as a regular attribute, use it in the op body. | ✅ confirmed | Probe passes; both TP2 tests pass; non-distributed `test_basic_model_load` still passes. | `logs/tp2-matches-merge/post-fix.txt` |

## Experiment log

*Reverse-chronological.*

<details>
<summary><b>20:13</b> — fix landed and validated</summary>

```
uv run pytest tests/test_distributed_tp2.py -m "not upstream" --tb=short
2 passed, 467 deselected in 274s

uv run pytest tests/test_vllm_spyre_next.py -m "not upstream" -k test_basic_model_load --tb=short
1 passed, 467 deselected in 69s
```

</details>

<details>
<summary><b>20:08</b> — applied fix in <code>spyre_inference/custom_ops/rotary_embedding.py</code></summary>

```diff
 # __init__:
+self._cpu_cos_sin_cache = self.cos_sin_cache
 # (and dropped the now-useless _apply override)

 # _rotary_cpu_op_func:
-cos_sin_cache=layer.cos_sin_cache,
+cos_sin_cache=layer._cpu_cos_sin_cache,
```

</details>

<details>
<summary><b>20:00</b> — root cause identified via traced <code>_buffers.__setitem__</code> hook</summary>

```
(Worker) [ROTARY BUFSET] cos_sin_cache: cpu -> spyre:0
   File "spyre_inference/v1/worker/spyre_model_runner.py", line 267, in load_model
     self.model.to(device=self._spyre_device)
   File "torch_spyre/model_utils.py", line 254, in _spyre_module_to
```

</details>

- **19:54** — proved `_apply` override never fires.
- **19:50** — proved `_match_cos_sin_cache_dtype` never fires (rules out the obvious upstream rotary path).
- **19:46** — baseline captured: `NotImplementedError: aten::index_select on spyre` from `RotaryEmbedding.forward_static` at `base.py:152`.

## Cluster validation

| Node id | Pre-fix | Post-fix | Notes |
|---|---|---|---|
| `test_tp2_llm_generate_matches_tp1` | ❌ FAIL (NotImplementedError on warmup) | ✅ **PASS** (213s) | Probe; both TP=1 and TP=2 sub-runs complete and the prefix-match assertion passes. |
| `test_tp2_llm_construction` | ❌ (implicit FAIL — same warmup path) | ✅ **PASS** | Same shared warmup path; ran in same session as `matches` for total 274s. |
| `test_basic_model_load` | (passing on main, would have regressed) | ✅ **PASS** (69s) | Non-distributed sanity for the rotary path. |

## Key findings / artifacts

- **Bug class:** external dependency monkey-patches a hook (`nn.Module.to`) and bypasses the abstraction we were relying on (`_apply`). This is squarely the kind of "torch-spyre changed underneath us" failure CLAUDE.md and the `debug-spyre` skill warn about. The signal that pointed at the right answer fastest was the stack at `_buffers.__setitem__` — once we instrumented the dict, the rest fell out in one run.
- **Files modified (final):** `spyre_inference/custom_ops/rotary_embedding.py` only. ~6 lines net.
- **Wasted hypotheses:** #2 (`_match_cos_sin_cache_dtype`) and #3 (`_apply`) — both reasonable starts but neither survives the trace. In hindsight, since the `_apply` tracer already showed the override was never fired, hypothesis #5 (instrument `_buffers` directly) should have come immediately after #3 instead of #4 (setattr).
- **Session dir:** `.claude/skills/debug-spyre/logs/tp2-matches-merge/` — `baseline.txt`, `debug-print.txt`, `trace-match.txt`, `trace-apply-stack.txt`, `trace-setattr.txt`, `trace-bufset.txt`, `post-fix.txt`.

## Retrospective

The real cost was discovering that `nn.Module.to` had been monkey-patched by torch-spyre. `CLAUDE.md` flags torch-spyre's `_monkey_patch.py` as a place to read, but this patch lives in `model_utils.py` — worth adding to the canonical list. A grep over `nn.Module.to =` in site-packages would have surfaced it in the first 30 seconds.

The fix is conservative: it adds an attribute, doesn't change any control flow. If torch-spyre later starts respecting `_apply` again (unlikely, but possible), the saved CPU reference is still correct — it's just slightly redundant. No need to revisit.

joerunde:
bot:next-test
export TEST_IMAGE_TAG=dev-2026_06_22-130225

joerunde:
bot:next-test
export TEST_IMAGE_TAG=dev-2026_06_22-130225

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -124,9 +124,9 @@ def forward_oot(self, x: torch.Tensor, bias: torch.Tensor | None = None) -> torc
         """OOT forward pass — lm_head matmul on Spyre.
 
         Called by SpyreUnquantizedLMHeadMethod.apply() from within
-        LogitsProcessor._get_logits(). Converts x (arriving on cpu)
-        to the weight device (residing on spyre), runs the compiled F.linear on spyre
-        and converts back to the x device (cpu).
+        LogitsProcessor._get_logits(). Runs the compiled F.linear on spyre,
+        performs the tensor slicing operation on CPU and converts back
+        to the original x device (spyre).
```
tdoublep: This is a great example of churny comments. If we have comments explaining what the code does, they need to be changed every time the code changes. The code should ideally be self-explanatory and comments used only in cases that are particularly hard to understand.

spyre_inference/custom_ops/__init__.py
```diff
@@ -30,5 +31,5 @@
 @lru_cache(maxsize=1)
 def register_all():
     logger.info("Registering custom ops for spyre_inference")
+    utils.register()
     rotary_embedding.register()
-    rms_norm.register()
```
tdoublep, line 34: why do we remove RMS norm?

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -142,13 +142,10 @@ def forward_oot(self, x: torch.Tensor, bias: torch.Tensor | None = None) -> torc
...
 
-        # Due to indexing operations inside the ModelRunner, which have
-        # to be carried out on cpu due to a torch-spyre limitation,
-        # the input to the SpyreParallelLMHead resides on CPU.
-        # Due to a second limitation of torch-spyre regarding sizes that can be used
+        # Due to a limitation of torch-spyre regarding sizes that can be used
         # in a F.linear layer, the original weights need to be padded
         out = F.linear(
-            convert(x, device=self.weight.device),
+            x,
```
tdoublep: Is this change related to wrapping the convert, or it's something else?

spyre_inference/custom_ops/rms_norm.py
tdoublep, line 1: There are a lot of changes in this file. Are these changes strictly related to wrapper the convert into an op, or it is just general cleanup?

spyre_inference/custom_ops/rotary_embedding.py
```diff
@@ -12,82 +12,141 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Spyre OOT replacement for RotaryEmbedding (CPU fallback).
+"""Spyre OOT replacement for RotaryEmbedding.
 
-Remove this file once Spyre natively supports rotary embedding ops.
+The CPU body is wrapped in a `direct_register_custom_op` so torch.compile does
+NOT trace into RotaryEmbedding.forward_native.
```
tdoublep: Comment churn

spyre_inference/custom_ops/rotary_embedding.py
tdoublep, line 1: Same question on this file

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -84,7 +84,12 @@ def _overwrite(
     if output.device.type == "spyre":
         # `torch.ops.spyre.overwrite` is dynamically registered, so its
         # signature is opaque to the type checker (ParamSpec resolves to `...`).
-        torch.ops.spyre.overwrite(input, output, dims, offsets)  # ty: ignore[invalid-argument-type]
+        torch.ops.spyre.overwrite(
+            input,  # ty: ignore[invalid-argument-type]
+            output,  # ty: ignore[invalid-argument-type]
+            dims,  # ty: ignore[invalid-argument-type]
+            offsets,  # ty: ignore[invalid-argument-type]
+        )
```
tdoublep: why is this needed?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -163,6 +163,19 @@ def _to_cpu(x):
...
+        gpu_model_runner.execute_model slices `hidden_states[logits_indices]`
+        on CPU (Spyre cannot slice), so the tensor handed to compute_logits
+        is on CPU. Thus, convert here, perform the lm_head operation and
+        then convert the resulting logits back to CPU
+        for downstream sampling.
+        """
+        hidden_states = convert(hidden_states, device=self._spyre_device)
+        logits = self._model.compute_logits(hidden_states, *args, **kwargs)
+        return convert(logits, device="cpu")
+
```
tdoublep: Also don't understand why this is added as part of this PR

tdoublep (commented):
Wrapping the convert in a custom op looks great! I'm just a bit unclear on whether all the other code changes are strictly related to that or are separate cleanup. Maybe I've missed the connection.

spyre_inference/custom_ops/utils.py
```diff
@@ -88,7 +84,58 @@ def convert(tensor, device=None, dtype=None):
...
+@lru_cache(maxsize=1)
+def register():
+    """Register the spyre_convert custom op with vLLM."""
+    # CompositeExplicitAutograd so the op dispatches regardless of input device
+    # (convert is called with both CPU and Spyre input tensors).
+    direct_register_custom_op(
+        op_name="spyre_convert",
+        op_func=_convert_op_func,
+        fake_impl=_convert_op_fake,
+        dispatch_key="CompositeExplicitAutograd",
```
tdoublep, line 144: Why is the dispatch key set this way?

tdoublep (approved):

spyre_inference/custom_ops/utils.py
bohnstingl: The convert function is being invoked in two scenarios:

1. Where the input is on CPU -> The transfer from CPU to Spyre.
2. Where the input is on Spyre -> The transfer from Spyre to CPU.

Thus, we can either register it twice: Once under `CPU` and once under `PrivateUse1`, or just use the `CompositeExplicitAutograd` dispatch key, which covers both.

bohnstingl:
@tdoublep you are rightfully pushing back. The initial form of this PR wasn't clean and it mixed reworks of the CustomOps with the opaque convert wrapper. Thus, I have split this PR into two:

- This PR, solely focusing on the opaque convert wrapper and 
- A follow-up PR one for the CustomOps rework, see https://github.com/torch-spyre/spyre-inference/pull/310

joerunde (approved):
lpgtm!

PR #260, opened 2026-06-09
Enable torch compile

### LINKED ISSUE #245: [Feature]: Turn off forced eager-mode.
### Feature description

With this issue resolved, the developer can choose to run in eager or in torch.compile mode.

### Motivation and context

This is the final piece to enable torch.compile in spyre-inference, see https://github.com/torch-spyre/spyre-inference/issues/243

### Proposed solution

_No response_

### Checklist

bohnstingl:
## Description

This PR removes the hard-coded enforce_eager=True from the platform.py, see https://github.com/torch-spyre/spyre-inference/blob/54fafa01df794217388dbc25a1a1781ce96b2dd6/spyre_inference/platform.py#L141-L146
With this, the user can choose to use torch.compile or eager-mode to execute the models. However, there is currently a torch-spyre issue blocking this PR from landing. We will open the respective issue and PR soon.

## Related Issues

https://github.com/torch-spyre/spyre-inference/issues/245

## Test Plan

TBD

## Checklist

### DIFF
--- a/spyre_inference/custom_ops/parallel_lm_head.py
+++ b/spyre_inference/custom_ops/parallel_lm_head.py
@@ -124,9 +124,9 @@ class SpyreParallelLMHead(ParallelLMHead):
         """OOT forward pass — lm_head matmul on Spyre.
 
         Called by SpyreUnquantizedLMHeadMethod.apply() from within
-        LogitsProcessor._get_logits(). Converts x (arriving on cpu)
-        to the weight device (residing on spyre), runs the compiled F.linear on spyre
-        and converts back to the x device (cpu).
+        LogitsProcessor._get_logits(). Runs the compiled F.linear on spyre,
+        performs the tensor slicing operation on CPU and converts back
+        to the original x device (spyre).
 
         At TP>1: Each rank computes logits for its vocabulary shard. The weight
         matrix is [num_embeddings_per_partition, hidden_dim], producing logits
@@ -142,13 +142,10 @@ class SpyreParallelLMHead(ParallelLMHead):
         """
         x_device = x.device
 
-        # Due to indexing operations inside the ModelRunner, which have
-        # to be carried out on cpu due to a torch-spyre limitation,
-        # the input to the SpyreParallelLMHead resides on CPU.
-        # Due to a second limitation of torch-spyre regarding sizes that can be used
+        # Due to a limitation of torch-spyre regarding sizes that can be used
         # in a F.linear layer, the original weights need to be padded
         out = F.linear(
-            convert(x, device=self.weight.device),
+            x,
             self.padded_weight.data,
             bias,
         )
--- a/spyre_inference/custom_ops/rms_norm.py
+++ b/spyre_inference/custom_ops/rms_norm.py
@@ -12,83 +12,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Spyre-specific RMSNorm implementation using out-of-tree (OOT) registration.
+"""Spyre OOT replacement for RMSNorm.
 
-This module provides a custom RMSNorm layer for IBM's Spyre device,
-replacing the upstream vLLM implementation (vllm/model_executor/layers/layernorm.py)
-when instantiated.
+Inputs and outputs both reside on Spyre, so forward_oot is fully traceable
+by Dynamo and no custom op boundary is needed. The kernel itself is wrapped
+in _maybe_compile so it is compiled independently when called eagerly.
 
-Architecture:
-    - OOT Registration: @RMSNorm.register_oot() replaces upstream at instantiation
-    - forward_oot(): Entry point for OOT dispatch, calls custom op for
-      torch.compile opacity
-    - Custom Op Boundary: torch.ops.vllm.spyre_rmsnorm is opaque to torch.compile,
-      so _forward_spyre_impl runs eagerly outside the compiled graph
-    - Separate Compilation: forward_spyre is compiled independently via maybe_compile
-
-Spyre Device Constraints:
-    - Minimum batch size: 64 (due to spyre constraint, automatically padded)
-    - Computations performed in torch.float16:
-      Input (dtype defined by model / user) converted to torch.float16 for
-      operations on spyre and then converted back to original dtype for cpu.
-    - Epsilon as tensor: Instead of a scalar, a tensor is created via torch.full()
-
-Limitations:
-    Currently the implementation in `forward_spyre` is similar to the
-    upstream implementation in `forward_static` from vllm/model_executor/layers/layernorm.py,
-    but it DOES NOT use the promotion of the data types, as this is not
-    yet supported in torch-spyre.
+Spyre constraints:
+    - Computation is performed in torch.float16.
+    - No dtype promotion to float32 (not yet supported in torch-spyre), so
+      numerical results may differ slightly from upstream vLLM.
+    - variance_epsilon is materialized as a tensor (scalar broadcast unsupported).
 
 References:
     - Upstream RMSNorm: vllm/model_executor/layers/layernorm.py
 """
 
 import torch
-import torch.utils._pytree as pytree
 
 from vllm.logger import init_logger
-from vllm.utils.torch_utils import direct_register_custom_op
 from vllm.model_executor.layers.layernorm import RMSNorm
-from functools import lru_cache
-
-from .utils import convert, register_layer, get_layer, _fake_impl
 
 logger = init_logger(__name__)
 
 
 @RMSNorm.register_oot(name="RMSNorm")
 class SpyreRMSNorm(RMSNorm):
-    """Out-of-tree (OOT) RMSNorm implementation for IBM's Spyre device.
-
-    This replaces the upstream vLLM RMSNorm (vllm/model_executor/layers/layernorm.py)
-    when instantiated, providing Spyre-specific optimizations and device handling.
-    """
+    """OOT RMSNorm for IBM's Spyre device."""
 
     _dynamic_arg_dims = {"x": [], "residual": []}
 
     def __init__(self, *args, **kwargs):
-        """Initialize SpyreRMSNorm layer.
-
-        Compiles the Spyre kernel based on SPYRE_INFERENCE_RMSNORM_KERNEL
-        environment variable and registers this instance in static_forward_context.
-        """
         super().__init__(*args, **kwargs)
-
-        self._target_device = torch.device("spyre")
-        self._target_dtype = torch.float16
-        self.maybe_compiled_forward_spyre = self.maybe_compile(self.forward_spyre)
-
-        self._layer_name = register_layer(self, "spyre_rmsnorm")
+        self._compiled_forward_spyre = self.maybe_compile(self._forward_spyre_impl)
 
         logger.warning_once(
             "SpyreRMSNorm: no dtype promotion is performed, "
             "expect numerical differences to upstream vLLM."
-        )
-        logger.debug_once(
-            "SpyreRMSNorm: Dispatch: enabled=%s, Forward method=%s, Compiled=%s",
-            self.enabled(),
-            self._forward_method.__name__,
-            self.maybe_compiled_forward_spyre is not self.forward_spyre,
         )
 
     def forward_oot(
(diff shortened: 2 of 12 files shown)


bohnstingl:
Closed in favor of https://github.com/bohnstingl/spyre-inference/pull/3

bohnstingl:
I reworked this PR quite a bit. It is now standalone and does not require any other PR targeting spyre-inference to land.
Still it relies on https://github.com/torch-spyre/torch-spyre/pull/3048 on the torch-spyre side to land. 

I tested this PR locally with our E2E example for Granite3.3-8B, as well as I ran all unit tests locally.

spyre_inference/custom_ops/utils.py
```diff
@@ -75,7 +75,10 @@ def _convert_op_func(
     target_dtype = dtype if dtype is not None else tensor.dtype
 
     if tensor.device.type == target_device.type and tensor.dtype == target_dtype:
-        return tensor
+        # Same-device, same-dtype: return FRESH, due to current torch-spyre limitatins.
```
jvlunteren: typo: limitatins

spyre_inference/platform.py
```diff
@@ -125,14 +125,18 @@ def apply_config_platform_defaults(cls, vllm_config: VllmConfig) -> None:
...
-
-        # Force eager execution. torch.compile with the Spyre inductor
-        # backend requires ALL graph tensors on Spyre, but our CPU fallback
-        # ops (embedding, linear, rotary, attention) create intermediate
-        # CPU tensors that the Spyre backend cannot codegen. Once all layers
-        # run natively on Spyre, this can be removed to enable compilation.
-        vllm_config.model_config.enforce_eager = True
+        # When enforce_eager is set, vLLM has already reset the mode to NONE;
+        # preserve that so eager stays eager.
+        if vllm_config.model_config is not None and vllm_config.model_config.enforce_eager:
```
jvlunteren: I think that `apply_config_platform_defaults` is called after `model_config` is constructed. Therefore, `vllm_config.model_config` is never `None` at this point. The `is not None` guard is dead code in that case.

spyre_inference/custom_ops/rms_norm.py
```diff
@@ -49,39 +46,26 @@ def forward_oot(
...
+            )
 
         variance = x.pow(2).mean(dim=-1, keepdim=True)
 
-        x = x * torch.rsqrt(variance + variance_epsilon)
+        x = x * torch.rsqrt(variance + self.variance_epsilon)
 
-        if weight is not None:
-            x = x * weight
+        if self.weight.data is not None:
```
jvlunteren: `self.weight` is a `torch.nn.Parameter`, so self.weight.data is always a Tensor and never `None`. The guard always evaluates true, which means when `self.has_weight` is `False`, the layer still runs `x = x * self.weight` against a zero-filled or uninitialized tensor, silently corrupting the output.

Fix:
```
if self.has_weight:
    x = x * self.weight
```

spyre_inference/platform.py
bohnstingl: So Claude tells me

The code
`platform.py:130` inside `apply_config_platform_defaults`:

```
if vllm_config.model_config is not None and vllm_config.model_config.enforce_eager:
    vllm_config.compilation_config.mode = CompilationMode.NONE
...
```

**What vLLM actually guarantees**
The hook is called from VllmConfig.__post_init__ at `vllm/config/vllm.py:1133`. Two facts contradict the reviewer's premise:

model_config: ModelConfig = None — its declared default is None (vllm.py:297).
vLLM's own __post_init__ guards self.model_config is not None at nearly every access — lines 865, 910, 917, 1027, 1037, 1050, and notably 1061, which is the near-identical enforce_eager check just before it calls our hook. If vLLM guaranteed non-None here, all those guards would be dead too.
So it is not true that model_config can never be None at this point.

But removing it is still correct — for a different reason
Our own code already dereferences model_config unconditionally on platform.py:144(vllm_config.model_config.dtype = ...). So if model_config were ever None, line 144 would crash regardless. The guard on line 130 therefore protects nothing — it just makes the code look defensive while the very next unconditional access defeats it.

So the guard is effectively dead, and removing it is fine:
```

I thus removed it.

File: spyre_inference/custom_ops/utils.py (hunk 1 above)
Comment (bohnstingl) [reply spyre_inference/custom_ops/utils.py]: Corrected.

File: spyre_inference/custom_ops/rms_norm.py (hunk 3 above)
Comment (bohnstingl) [reply spyre_inference/custom_ops/rms_norm.py]: Good catch. I missed this during my refactor. Corrected.

**Comment by bohnstingl** [comment]
FYI, the torch-spyre PR https://github.com/torch-spyre/torch-spyre/pull/3048 this work depends on has landed in `main`.

File: spyre_inference/custom_ops/utils.py [hunk 4]
```diff
@@ -75,7 +75,10 @@ def _convert_op_func(
     target_dtype = dtype if dtype is not None else tensor.dtype
 
     if tensor.device.type == target_device.type and tensor.dtype == target_dtype:
-        return tensor
+        # Same-device, same-dtype: return FRESH, due to current torch-spyre limitations.
+        if tensor.device.type == "spyre":
+            return tensor.to("cpu").to(tensor.device)
+        return tensor.clone()
```
tdoublep: If the device is Spyre we move it back to CPU and then back to Spyre again? And if the device isn't Spyre we clone it? This seems like kind of a big chance that might impact perf

spyre_inference/custom_ops/utils.py
```diff
@@ -120,6 +123,12 @@ def convert(tensor, device=None, dtype=None):
         return None
     if isinstance(device, str):
         device = torch.device(device)
+    # Short-circuit a true no-op at the call site so Inductor never emits a
+    # same-device/dtype spyre_convert FallbackKernel into the graph.
+    target_device = device if device is not None else tensor.device
+    target_dtype = dtype if dtype is not None else tensor.dtype
+    if tensor.device.type == target_device.type and tensor.dtype == target_dtype:
+        return tensor
```
tdoublep, line 100: Oh I see, you kind of move it earlier to avoid Inductor issue

spyre_inference/custom_ops/utils.py
tdoublep: I think I understand having read the code below. Do we ever expect this path to get triggered though? Could just assert that it never happens?

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -42,23 +44,30 @@ def __init__(self, *args, **kwargs):
...
             )
-            masked_input = convert(masked_input, device=input_.device)
+            masked_input = convert(masked_input, device=self.weight.data.device)
         else:
             masked_input = input_
+            input_mask = None
 
-        output_parallel = self.quant_method.embedding(self, masked_input.long())
+        weight = cast(torch.Tensor, self.weight.data)
+        output = torch.nn.functional.embedding(masked_input, weight)
```
tdoublep: Why do we no longer call `self.quant_method.embedding` ?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -294,41 +295,40 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
...
             self.model,
             backend="inductor",
             fullgraph=True,
             dynamic=False,
         )
-        logger.info("Model compiled for Spyre (backend=inductor) in %.3fs.", time.time() - t0)
+        logger.info(
+            "Compiled model %s as a single graph for Spyre in %.3fs.",
+            type(self.get_model()).__name__,
+            time.time() - t0,
```
tdoublep, line 456: Is this really where the compilation happens? Isn't it lazy?

spyre_inference/custom_ops/utils.py
bohnstingl: Indeed this is a very corner case that we should avoid in general. We now raise an Exception if that were to happen.

spyre_inference/custom_ops/vocab_parallel_embedding.py
bohnstingl: This was an inadvertent change. Reverted.

bohnstingl:
I updated the PR and had to rework the spyre_communicator for this as well. In particular, we have to utilize the gloo-path for cross-device communication, which means `spyre tensor -> D2H -> gloo collective -> H2D -> spyre tensor`.

I also traced down the current failure for the transformer model. It is cause because the Granite model in the RMSNorm (`GraniteRMSNorm`) does a dtype upcast to FP32, see https://github.com/huggingface/transformers/blob/150eb7c9ed4091294c829fa0e9466b090cb0f87f/src/transformers/models/granite/modeling_granite.py#L194. This causes issue with torch.compile. There are a couple of things to be investigated:

1. Is the dtype upcast in RMSNorm properly supported even in torch-spyre itself?
2. If 1.) is supposed to work, investigate why it fails under torch.compile?
3. Can we find an alternative transformer model that does not do this?
4. Shall we xfail the transformer tests for the moment? This is my least preferable option tbh.

bohnstingl:
For the moment, are xfailing the transformer test because of the unsupported `dtype` upcast in torch.compile.
I am exploring this as a separate thread together with the rework of the `rms_norm` more generally.

tests/test_spyre_comms_native_probes.py
```diff
@@ -107,5 +107,15 @@ def test_native_all_gather_list_works(run_tp_probe) -> None:
     _spyre_device_count() < 2,
     reason="needs >=2 Spyre cards; skipping TP=2 native-probe test",
 )
+@pytest.mark.xfail(
```
joerunde: why was this added?

If there is a regression with any of the native comms functionality that we need, we should pinpoint the library version that introduces it and open an issue

tests/test_transformers_backend.py
```diff
@@ -24,6 +24,17 @@
...
+# RMSNorm/LayerNorm upcast to float32 inside the norm
+# (`hidden_states.to(torch.float32)`), which the torch-spyre Inductor backend
+# cannot lower: the subsequent multi-arg pointwise mixes the fp32-converted
+# (DL16_TO_FP32 "staggered" ElementArrangement) operand with a full-width
+# STANDARD operand, raising `Unsupported: Multi-arg pointwise with mixed EA ...`.
+@pytest.mark.xfail(
+    reason="transformers backend does not compile on Spyre:\
+        unsupported dtype upcast under torch.compile",
+    strict=True,
+)
```
joerunde: Shouldn't we disable compilation on the transformers backend instead of xfailing the test?

bohnstingl:
Some updates:

- fullgraph compile is currently disabled by default, but it would be supported if turned on. There is an additional small testcase that compiles a model and verifies its output. 
- @yannicks1, I reworked the RoPE implementation a bit. In particular, there is now another opaque wrapper to hide the math from dynamo itself and the constants used in the implementation are lifted and computed once. WDYT?


The current PR has the following limitations:

1. It relies on https://github.com/torch-spyre/torch-spyre/pull/3323
2. We have to wrap the RoPE implementation into an opaque wrapper so that Dynamo doesn't trace it. This is because of a torch-spyre issues that needs to be investigated
3. Nested-compilation does not work for RoPE, because of the mixed CPU-spyre graphs.

For this PR, I will next address all the CI fails and introduce a comparison between eager and compile-mode in the test.

As a follow-up, I am investigating point 2 and 3 above, as these might have broader implications beyond RoPE

tests/test_compile.py
```diff
@@ -0,0 +1,55 @@
...
+        "ibm-ai-platform/micro-g3.3-8b-instruct-1b",
+        pytest.param(
+            "google/gemma-3-1b-it",
+            marks=pytest.mark.skip(reason="Gemma3 currently doesn't work with torch.compile"),
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "ref_output",
+    ["\n\nA list of Identified Benefits under Debt Management – Count", "\n\nIBM's main"],
```
joerunde: @bohnstingl this doesn't make sense to me- why would there be two separate reference outputs for the same prompt? Only one can possibly match

If this was meant to be one reference for `ibm-ai-platform/micro-g3.3-8b-instruct-1b` and one for `google/gemma-3-1b-it` then the parameters need to be specified differently. As is this generates the cross product of parameterizations:
```
tests.test_compile
  ✅ test_basic_llm_inference[\n\nA list of Identified Benefits under Debt Management \u2013 Count-ibm-ai-platform/micro-g3.3-8b-instruct-1b]
  ⚪ test_basic_llm_inference[\n\nA list of Identified Benefits under Debt Management \u2013 Count-google/gemma-3-1b-it]
  ❌ test_basic_llm_inference[\n\nIBM's main-ibm-ai-platform/micro-g3.3-8b-instruct-1b]
	AssertionError: Model produced wrong output! 
	  - A list of Identified Benefits under Debt Management – Count
	  + IBM's main
  ⚪ test_basic_llm_inference[\n\nIBM's main-google/gemma-3-1b-it]
```

spyre_inference/custom_ops/parallel_lm_head.py
```diff
@@ -63,7 +62,10 @@ def process_weights_after_loading(self, layer):
                 layer.padded_weight.shape[0],
             )
         else:
-            layer.padded_weight = layer.weight
+            # Clone → INDEPENDENT storage from `weight`. With
+            # tie_word_embeddings, `weight` IS `embed_tokens.weight` and must
+            # stay on CPU; `_apply` below moves only `padded_weight` to Spyre.
```
joerunde: Which `_apply` does this comment apply to?

spyre_inference/custom_ops/utils.py
```diff
@@ -44,7 +44,10 @@ def _convert_op_func(
     target_dtype = dtype if dtype is not None else tensor.dtype
 
     if tensor.device.type == target_device.type and tensor.dtype == target_dtype:
-        return tensor
+        raise RuntimeError(
+            f"Trying to convert a tensor to the same device ({tensor.device.type}) "
+            + "and same dtype ({tensor.dtype}), should never happen!"
```
joerunde: needs to be an f string here to format `{tensor.dtype}`

spyre_inference/platform.py
```diff
@@ -182,14 +182,23 @@ def apply_config_platform_defaults(cls, vllm_config: VllmConfig) -> None:
...
+        # When enforce_eager is set, vLLM has already reset the mode to NONE;
+        # preserve that so eager stays eager.
+        # NOTE: If vllm_config.compilation_config.mode is None and
+        # vllm_config.model_config.enforce_eager == False,
+        # no particular compilation mode has been selected. Continue in eager for the moment
+        if vllm_config.model_config.enforce_eager or vllm_config.compilation_config.mode is None:
+            vllm_config.compilation_config.mode = CompilationMode.NONE
+        else:
+            # Only if enforce_eager=False and a particular CompilationMode is selected,
+            # continue in compile mode
```
joerunde, line 204: do we want a `log.warning_once` if the user specified a different compilation mode, like `VLLM_COMPILE` or `DYNAMO_TRACE_ONCE`, so the user knows their selection was overridden?

tests/test_compile.py
```diff
@@ -0,0 +1,55 @@
...
+# limitations under the License.
+
+"""Torch.compile tests"""
+
+from __future__ import annotations
+
+import pytest
+
+
+@pytest.mark.compile
```
joerunde: new markers should go into pyproject.toml- but do we need a new marker here if this is the only test?

could consider adding later if/when we have more tests and want to split out a CI job for just the full graph compile tests

tests/test_compile.py
bohnstingl: This should be resolved now

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: I fixed the comment.

spyre_inference/custom_ops/utils.py
bohnstingl: Fixed.

spyre_inference/platform.py
bohnstingl: I added a warning.

tests/test_compile.py
bohnstingl: I removed the marker for the moment. As suggested, we can put it back at a later time when needed.

PR #262, opened 2026-06-10
Enforce block_size % 64 == 0

### LINKED ISSUE #239: [Bug]: Attention CI tests fail with small block_sizes
### Describe the bug

The [list-based Spyre attention backend](https://github.com/bringlein/spyre-inference/blob/4c10b52db296b710dad51e5d7dd9ddc147912c3f/spyre_inference/v1/attention/backends/spyre_attn.py) crashes the Spyre inductor's layout propagator when KV pages are allocated with a `block_size` different from `head_size`. The crash happens inside the per-page online-softmax kernel, on the first matmul that consumes a transposed K page.

**Failure signature** (reproduces on the current branch with `tests/test_spyre_attn.py` parametrized with `block_size=16, head_size=128`):

```
torch._inductor.exc.InductorError: Unsupported: Spyre backend does not support:
  mutation op buf0 arg=arg4_1: cannot restickify [128, 64, 2048, 1] -> [2048, 16, 64, 8192, 1]
at torch-spyre/torch_spyre/_inductor/insert_restickify.py:321
```

**Where it originates** — `_create_compilable_page_attn` in `spyre_inference/v1/attention/backends/spyre_attn.py`, the K-side matmul:

```python
scores = _indirect_matmul_mock(
    q, None, k_pages, page_idx,
    transform_b=lambda t: t.unsqueeze(1).transpose(-2, -1),
)
```

`k_pages[idx]` has shape `[num_kv_heads, block_size, head_size]`. The view `unsqueeze(1).transpose(-2, -1)` produces `[num_kv_heads, 1, head_size, block_size]`. When `block_size == head_size` (the production case for Granite-3.3-8B: both are 128) the resulting Spyre stick layout is restickifiable; when they differ, the propagator can't construct the target layout `[2048, 16, 64, 8192, 1]` from the source `[128, 64, 2048, 1]` and raises.

This issue is related and a follow-up to https://github.com/torch-spyre/spyre-inference/pull/206

### How to reproduce

Run the current attention tests with a `block_size` different than the `head_size`, i.e., `16`.

### Additional context

_No response_

### Checklist

- [x] I have searched for similar issues.

rafvasq:
## Description

Enforces `block_size % 64 == 0` for the list-based Spyre attention backend

- Round default `block_size` up to the nearest multiple of 64 in `check_and_update_config`
- Raise `ValueError` for explicitly set invalid values
- xfail upstream test `test_causal_backend_correctness` (uses `block_size=16`)
- Add unit tests for the backend validation and the platform override

## Related Issues

Addresses #239, more discussion there on the investigation and approach

## Test Plan

- `tests/test_spyre_attn.py` includes block-size validation
- `tests/test_platform.py` includes block-size override cases

## Checklist

- [] I have updated the documentation (if applicable)

### DIFF
--- a/spyre_inference/v1/attention/backends/spyre_attn.py
+++ b/spyre_inference/v1/attention/backends/spyre_attn.py
@@ -342,6 +342,14 @@ class SpyreAttentionMetadataBuilder(AttentionMetadataBuilder[SpyreAttentionMetad
         self.block_size = kv_cache_spec.block_size
         self.head_size = kv_cache_spec.head_size
 
+        # Validate block_size alignment: Spyre stick size is 128 bytes (64 fp16 elements).
+        # block_size must be a multiple of 64 to avoid restickification errors
+        if self.block_size % 64 != 0:
+            raise ValueError(
+                f"block_size ({self.block_size}) must be a multiple of 64 for Spyre "
+                f"Got block_size={self.block_size}, head_size={self.head_size}."
+            )
+
         model_config = vllm_config.model_config
         self.num_heads = model_config.get_num_attention_heads(vllm_config.parallel_config)
         self.num_kv_heads = model_config.get_num_kv_heads(vllm_config.parallel_config)
@@ -499,7 +507,10 @@ class SpyreAttentionBackend(AttentionBackend):
 
     @staticmethod
     def get_supported_kernel_block_sizes() -> list[int | MultipleOf]:
-        return [MultipleOf(1)]
+        # Spyre stick size is 128 bytes; tensors are transferred as float16 (2 bytes),
+        # so block_size must be a multiple of 64 (= 128 / 2) to satisfy stick alignment.
+        # This matches the constraint on head_size in supports_head_size().
+        return [MultipleOf(64)]
 
     @staticmethod
     def get_name() -> str:
--- a/spyre_inference/v1/attention/backends/spyre_attn.py
+++ b/spyre_inference/v1/attention/backends/spyre_attn.py
@@ -343,11 +343,12 @@ class SpyreAttentionMetadataBuilder(AttentionMetadataBuilder[SpyreAttentionMetad
         self.head_size = kv_cache_spec.head_size
 
         # Validate block_size alignment: Spyre stick size is 128 bytes (64 fp16 elements).
-        # block_size must be a multiple of 64 to avoid restickification errors
+        # block_size must be a multiple of 64 to avoid restickification errors during
+        # torch.compile.
         if self.block_size % 64 != 0:
             raise ValueError(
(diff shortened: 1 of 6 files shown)


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -375,14 +342,19 @@ def __init__(
...
         self.head_size = kv_cache_spec.head_size
 
+        # Validate block_size alignment: Spyre stick size is 128 bytes (64 fp16 elements).
+        # block_size must be a multiple of 64 to avoid restickification errors during
+        # torch.compile.
+        if self.block_size % 64 != 0:
+            raise ValueError(
+                f"block_size must be a multiple of 64 for the list-based attention "
+                f"backend. Got block_size={self.block_size}, head_size={self.head_size}. "
+            )
```
joerunde, line 352: this is nice!

But should we put an override in platform.py so that we can ensure the block size is always a multiple of 64? Something quick and dirty like this would work:
```
if config.block_size % 64 != 0:
    logger.warning("Block size must be a multiple of 64, overriding %d to 64", config.block_size)
    config.block_size = 64
```

rafvasq: Added this in check_and_update_config to round up "non-user-specified" block sizes to the nearest multiple of 64 and left the backend ValueError in as a guard for explicitly-set invalid values, though I could have it override those too if we think it should always just work? (Also added related tests in test_platform.py.)

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -74,7 +74,9 @@ tests:
     - rel_path: tests/v1/attention/test_attention_backends.py
       allow_list:
         - test: "test_causal_backend_correctness"
-          mode: mandatory_pass
+          # Xfailed because the list-based backend requires block_size % 64 == 0
+          # for Spyre stick alignment. The test uses block_size=16.
```
joerunde: Can we instead put in a custom fixture or something to modify the test to use block_size==64?

The attention tests are important- we shouldn't just disable them

spyre_inference/v1/attention/backends/spyre_attn.py
joerunde: I could go either way- since this is only used after configuring via platform.py then we shouldn't need the extra check, but since tests can call the backend directly it is nice to know that tests are configured correctly too

joerunde:
This looks good to me, and the CI failures are unrelated (currently have a base image issue affecting the TP=2 tests)

joerunde (approved):

PR #264, opened 2026-06-10
Add Spyre paged KV connector NIXL path

toddllm:
Port the Spyre InMemory KV connector surface into spyre_inference, including environment knobs, factory registration, worker registration, and the worker-side connector bridge.

Add a paged KV cache accessor for the Spyre list-of-pages layout while keeping the heap accessor as the fallback for older or unknown cache layouts.

Add a connector-level NIXL smoke harness and focused tests for registration, structure, paged cache access, and split prefill/decode connector behavior.

## Checklist


toddllm:
This is not intended to be a single big PR. Just a saved state that we are testing against.

I'll break it up if we decide we want to include any of this.

PR #265, opened 2026-06-11
Add Spyre paged KV connector prototype with NIXL transfer

toddllm:
Adds a draft Spyre paged KV connector path for vLLM V1 KV transfer, including InMemorySpyreConnector, paged KV cache access, NIXL/UCX transfer, config-driven role/use_nixl/remote endpoint resolution, env override compatibility, connector smoke coverage, structural tests, and user-facing connector config docs.

Local validation: compileall, ruff, connector pytest set, and ty pass with 0 errors. Connector-level NIXL and env-driven two-pod serve have been validated separately. Config-driven two-pod serve validation is the active final gate and should be added to this draft once the IBM-side report lands.

Draft scope: not a production llm-d routing implementation, not performance tuned, and not yet a full routing/proxy validation.


toddllm:
Again, just POC at this point, but we are testing it. I'll do proper smaller PRs if any of this code is wanted

PR #270, opened 2026-06-11
feat: expand upstream test coverage with model e2e

kmehant:
This PR attempts to add some more tests from upstream vLLM targetting e2e support testing. For now unsupported tests are marked `skip`

### DIFF
--- a/tests/plugin/spyre_testing_plugin/upstream_tests.yaml
+++ b/tests/plugin/spyre_testing_plugin/upstream_tests.yaml
@@ -27,19 +27,19 @@ tests:
   files:
     # RMSNorm tests require either https://github.com/vllm-project/vllm/pull/36246 or vLLM IR
     # changes to merge in order to pass
-    # - rel_path: tests/kernels/core/test_layernorm.py
-    #   allow_list:
-    #     - test: "test_rms_norm"
-    #       mode: mandatory_pass
-    #       tags: [rmsnorm, upstream]
-    #       params:
-    #         skip:
-    #           strided_input: [true]
-    #         override:
-    #           num_tokens: [1, 16]
-    #           hidden_size: [64]
-    #   block_list:
-    #     - test: "test_fused_rms_norm_quant"
+    - rel_path: tests/kernels/core/test_layernorm.py
+      allow_list:
+        - test: "test_rms_norm"
+          mode: xfail
+          tags: [rmsnorm, upstream]
+          params:
+            skip:
+              strided_input: [true]
+            override:
+              num_tokens: [1, 16]
+              hidden_size: [64]
+      block_list:
+        - test: "test_fused_rms_norm_quant"
     - rel_path: tests/kernels/core/test_activation.py
       allow_list:
         - test: "test_act_and_mul"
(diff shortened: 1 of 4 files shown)


ashokponkumar (approved):

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -59,30 +59,98 @@ tests:
...
+        - test: "test_models"
+          mode: xfail
+          # uses_subprocess marker is added to make these tests run first.
+          # This must be added for any sets of tests that use `vllm.LLM`
+          tags: [upstream, uses_subprocess]
+          params:
+            override:
+              model:
+                - facebook/opt-125m
+                - ibm-granite/granite-3.3-8b-instruct
```
joerunde: This model generally isn't in the upstream test parameterizations, so where we've added it here we're just yolo'ing it in. That might not be the best idea because we don't even know that the test(s) will pass with this model on gpus.

Also, this is a pretty big model to be running with unit tests. There are other options like the granite micro model (`ibm-ai-platform/micro-g3.3-8b-instruct-1b`) that we can use to test the same granite architecture without needing to run a full-sized model. This is a concern even for mature accelerator support, but right now we're still swapping layers between cpu and spyre so this model would take _forever_ to run :(

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -59,30 +59,98 @@ tests:
...
-    #         allow:  # skip every model except facebook/opt-125m
-    #           model:
-    #           - facebook/opt-125m
+    - rel_path: tests/models/language/generation/test_common.py
+      allow_list:
+        - test: "test_models"
+          mode: xfail
+          # uses_subprocess marker is added to make these tests run first.
+          # This must be added for any sets of tests that use `vllm.LLM`
+          tags: [upstream, uses_subprocess]
```
joerunde: can we try to also set a descriptive marker that we can use to select different sets of tests? Maybe `model` for all the ones that run full models?

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -27,19 +27,19 @@ tests:
...
-    #       tags: [rmsnorm, upstream]
-    #       params:
-    #         skip:
-    #           strided_input: [true]
-    #         override:
-    #           num_tokens: [1, 16]
-    #           hidden_size: [64]
-    #   block_list:
-    #     - test: "test_fused_rms_norm_quant"
+    - rel_path: tests/kernels/core/test_layernorm.py
```
joerunde, line 30: The comment above this says that the upstream test needs to be changed (or vllm IR integration, which we paused at it was deprioritized by the vllm maintainers).

I'm not sure it makes sense to activate a test that we know is just incorrect and will always fail, I'd expect this to either stay commented or be deleted. (We can open up an issue to track the work to actually fix and enable this test, or explicitly decide as a team not to do it)

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -59,30 +59,98 @@ tests:
...
+              model:
+                [
+                  "Qwen/Qwen3-0.6B",
+                  "ibm-granite/granite-3.3-2b-instruct",
+                  "ibm-granite/granite-3.3-8b-instruct",
+                ]
+              backend: ["CUSTOM"]
+              enforce_eager: [true]
+        - test: "test_models_distributed"
+          mode: xfail
```
joerunde: Is it too wasteful of CI resources to enable all of these `xfail` tests?

We know there's work required to enable these so shouldn't we enable them as we do the work?

kmehant: I have marked them skip :)

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
kmehant: True, I have pushed for the PR to be merged irrespective of the vLLM IR work. However there was no response from the vLLM team. I have marked it skip for now.

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
kmehant: Have added a model marker.

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -80,15 +83,70 @@ tests:
...
+              model: ["Qwen/Qwen3-0.6B", "ibm-granite/granite-3.3-2b-instruct"]
+              backend: ["CUSTOM"]
+              enforce_eager: [true]
+        - test: "test_models_distributed"
+          mode: skip
+          tags:
+            [model, basic_correctness, upstream, uses_subprocess, distributed]
+          params:
+            override:
+              model: ["Qwen/Qwen3-0.6B", "ibm-granite/granite-3.3-2b-instruct"]
```
joerunde: just a style nit: can we have the params override lists match the formatting of the rest of the file, like
```
- Qwen/Qwen3-0.6B
- ibm-granite/granite-3.3-2b-instruct
```

kmehant: Done!

joerunde (approved):

PR #268, opened 2026-06-11
[Distributed] Remove gather override and update the test cases

R3hankhan123:
## Description

The latest spyre comms code adds native gather and list-form allgather to spyre_comms. The native probes test_native_gather_works and test_native_all_gather_list_works now pass, confirming the collectives work on a real spyreccl device_group at TP=2.

## Related Issues

Fixes #267 
## Test Plan
Run `test_spyre_comms_native_probes.py` file
```
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.12.13, pytest-9.0.2, pluggy-1.6.0 -- /home/rehankhan/dt-inductor/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/rehankhan/vllm-spyre-next/spyre-inference
configfile: pyproject.toml
plugins: anyio-4.13.0, xdist-3.8.0
collected 4 items                                                                                                                                                                            

tests/test_spyre_comms_native_probes.py::test_native_all_reduce_works XFAIL (libspyre_comms does not yet implement native allreduce; SpyreCommsContext::allreduce throws. When this
flips to passing, delete SpyreCommunicator.all_reduce.)                                                                                                                                [ 25%]
tests/test_spyre_comms_native_probes.py::test_native_all_gather_into_tensor_works XFAIL (torch-spyre's spyreccl backend stubs _allgather_base, so dist.all_gather_into_tensor fails
even though libspyre_comms implements single-tensor allgather. When this flips to passing, the base DeviceCommunicatorBase.all_gather can replace SpyreCommunicator.all_gather.)       [ 50%]
tests/test_spyre_comms_native_probes.py::test_native_all_gather_list_works PASSED                                                                                                      [ 75%]
tests/test_spyre_comms_native_probes.py::test_native_gather_works PASSED                                                                                                               [100%]

```

## Checklist

- [x] I have read the [contributing guidelines](https://torch-spyre.github.io/spyre-inference/contributing/)
- [x] My code follows the project's code style (run `bash format.sh`)
- [ ] I have added tests for my changes (if applicable)
- [ ] I have updated the documentation (if applicable)
- [x] My commits include a `Signed-off-by:` line (DCO compliance)

PR #271, opened 2026-06-11
🧪 Update stale test markers and run on spyre

joerunde:
## Description

Removes stale `xfail` markers on tests whose underlying limitations have already been fixed, and rewrites several tests so they actually exercise the Spyre device instead of running entirely on CPU.

**xfail removals**

- `test_silu_and_mul.py`: drops the `xfail(reason="Strided tensors on float16 does not work on spyre")` parametrization (36 XPASSes). The `forward_oot` path was reworked in #148 to slice on CPU and only push contiguous halves to Spyre, making the original Spyre-strided-fp16 limitation moot for this test. Also drops the redundant `float32` parametrization — the platform forces fp16, so fp32 wasn't representative of any production path.

**Tests rewritten to actually run on Spyre**

Several tests created Spyre OOT layers but then ran `forward()` against CPU weights and CPU inputs, so the math executed entirely on the host. They now move the layer to `"spyre"` and pass Spyre inputs where the production path does (CPU input where the layer is responsible for H2D):

- `test_mlp.py::test_merged_column_matches_reference` — Spyre weight + Spyre x.
- `test_mlp.py::test_qkv_matches_reference` — Spyre weight + Spyre x; asserts the result comes back on CPU (the D2H workaround in `SpyreQKVParallelLinear.forward`).
- `test_mlp.py::test_row_parallel_matches_reference` — Spyre weight + CPU x, exercising the H2D inside `SpyreRowParallelLinear.forward`.
- `test_parallel_lm_head.py::test_spyre_parallel_lm_head_matches_reference` — Spyre weight so `forward_oot` actually H2Ds and runs F.linear on Spyre. fp16 accumulation diverges from the CPU reference (already documented in `SpyreUnquantizedLMHeadMethod`'s "expect numerical differences" warning), so tolerances bumped to `atol=1e-1, rtol=5e-2`.
- `test_silu_and_mul.py::test_spyre_siluandmul_matches_reference` — pushes input to Spyre so the `if x.device.type == "spyre"` branch in `forward_oot` is what gets exercised.
- `test_vocab_parallel_embedding.py` (TP=1 and fake-TP=2): layers/inputs go to Spyre. `aten.embedding` currently has no Spyre kernel and falls back to CPU — left in place so this test automatically lights up real on-device coverage when the kernel lands; the comment calls out the current fallback.

Tests already exercising Spyre via custom-op dispatch (`test_rms_norm.py`) or device-parametrized fixtures (`test_spyre_attn.py`) are unchanged.

**Test marker fix**

`test_tp2_llm_construction` was tagged `@pytest.mark.spyre` instead of `@pytest.mark.distributed`, so it was skipped on the runner with a single card. Moved to `distributed` to match its sibling tests.

** Performance fix **

This turns down the kv cache size to the minimum required to run the model (max_num_seqs * max_model_len) to vastly speed up tests on CI. Otherwise we waste time allocating a huge kv cache that we don't use.

## Related Issues

N/A — cleanup of stale markers and test coverage gaps surfaced during the audit triggered by the XPASSes.

## Test Plan

```bash
uv run pytest -m "not upstream"

- Before: 209 passed, 8 skipped, 467 deselected, 8 xfailed, 36 xpassed in 472s
- After:  209 passed, 8 skipped, 467 deselected, 8 xfailed in 567s

The ~95s wall-clock increase is the rewritten tests now actually round-tripping through Spyre instead of staying on CPU. No new fallback warnings versus baseline (38 → 38).

Checklist

- [x] I have read the contributing guidelines
- [x] My code follows the project's code style (run bash format.sh)
- [x] I have added tests for my changes (if applicable) — this PR is test-only
- [ ] I have updated the documentation (if applicable) — N/A
- [x] My commits include a Signed-off-by: line (DCO compliance)


rafvasq (approved):
Went through this and confirmed most tests are actually running on spyre now and that the `aten.embedding` falls back to CPU. Also used claude to measure the actual divergence and confirmed the updated lm_head tolerances aren't overshooting. LGTM 👍

joerunde:
rebasing after #278, should unblock the tests here but still expecting one failure on `gather`

joerunde:
Failures related to gather- merging this now to cut the test time from ~30 minutes to ~7 and will shepherd in #268

PR #274, opened 2026-06-12
cicd: add spyre-inference integration dispatch

kmehant:
This PR adds ability to dispatch test workflows on desired torch-spyre and underneath dependency RPMs. This dispatch will be used as part of integration testing with upstream libraries.

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -128,11 +128,11 @@ jobs:
 
       - name: Build spyre-inference
         env:
-          INPUT_REF: "${{ inputs.torch_spyre_ref }}"
+          TORCH_SPYRE_REF: "${{ inputs.torch_spyre_ref }}"
         run: |
           source "$HOME/.bashrc"
           source /etc/profile.d/ibm-aiu-setup.sh
-          if [[ -n "$INPUT_REF" ]]; then
+          if [[ -n "$TORCH_SPYRE_REF" ]]; then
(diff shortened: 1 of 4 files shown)


ashokponkumar (approved):

.github/workflows/test_each_commit.yaml
```diff
@@ -99,14 +127,17 @@ jobs:
           echo "---"
 
       - name: Build spyre-inference
+        env:
+          INPUT_REF: "${{ inputs.torch_spyre_ref }}"
```
joerunde: nit: would be nice if this was more descriptive than just `INPUT_REF`

.github/workflows/test_each_commit.yaml
```diff
@@ -99,14 +127,17 @@ jobs:
...
-          uv sync \
-            --frozen \
-            --verbose \
-            --refresh \
-            --group dev
+          if [[ -n "$INPUT_REF" ]]; then
+            uv add torch-spyre /home/senuser/torch-spyre
+            uv sync --verbose --refresh --group dev
+          else
+            uv sync --frozen --verbose --refresh --group dev
```
joerunde: let's be sure to remove `--refresh` here when resolving the merge

joerunde (approved):
LPGTM- I assume this is here to allow different pieces of automation to trigger the workflow with newer torch-spyre versions for integration testing?

It would be great to add a description to the PR about this though, for future reference

Issue #72, opened 2026-02-19
[Epic] Support prefill/decode disaggregation


tdoublep:
We need the vLLM Spyre plugin to support P/D disaggregation for full integration with llm-d. 

This work will involve looking at the PoC developed by @fabianlim and team and integrating it properly into the vllm-spyre codebase.

fabianlim:
@tdoublep @Nhan-Hoang  can you add @toddllm to this issue as well.

PR #275, opened 2026-06-16
RFC: Port upstream NIXL Connector  to spyre inference

praveingk:
This Design document addresses https://github.com/torch-spyre/spyre-inference/issues/72

We propose to use the upstream NIXL Connector, and implement the Spyre-specific comments.

Initially, this RFC will enable PD by using a CPU buffer. This is an interim solution, until Spyre-to-Spyre communication path is well established. The current choice of using upstream NIXL Connector will easily transition to that using a Spyre-specific transport-backend.


praveingk:
@vknaik @toddllm

Issue #215, opened 2026-06-01
Add tp>1 for Linear Parallel LM Head


R3hankhan123:
Update linear parallel LM Head custom up to support TP>1

R3hankhan123:
Looks like numerical variance is there 
```
STDERR:
[rank1]: Traceback (most recent call last):
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 373, in <module>
[rank1]:     main()
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 367, in main
[rank1]:     PROBES[args.probe](device, device_group, world_size, rank)
[rank1]:   File "/home/rehankhan/vllm-spyre-next/spyre-inference/tests/probes/tp_probe.py", line 158, in probe_parallel_lm_head
[rank1]:     torch.testing.assert_close(out, expected, atol=1e-3, rtol=1e-3)
[rank1]:   File "/home/rehankhan/dt-inductor/.venv/lib64/python3.12/site-packages/torch/testing/_comparison.py", line 1600, in assert_close
[rank1]:     raise error_metas[0].to_error(msg)
[rank1]: AssertionError: Tensor-likes are not close!

[rank1]: Mismatched elements: 6719 / 32768 (20.5%)
[rank1]: Greatest absolute difference: 0.0078125 at index (14, 1618) (up to 0.001 allowed)
[rank1]: Greatest relative difference: 52.21875 at index (8, 994) (up to 0.001 allowed)
```

R3hankhan123:
Numerical variance is there in TP=1 also

PR #277, opened 2026-06-16
:bug: fix TP2 failure with cpu bounce, reinstate probes

### LINKED ISSUE #276: [Bug]: Regression in TP
### Describe the bug

Refer to this workflow run - https://github.com/torch-spyre/spyre-inference/actions/runs/27416897901/job/81043006924. There is regression in tensor parallelism (tp 1 and 2 do not provide same tokens out). This regression mostly coming from internal libs that needs identifying the part causing this. 

### How to reproduce

You can run the test `tests/test_distributed_tp2.py::test_tp2_llm_generate_matches_tp1`

### Additional context

_No response_

### Checklist

- [x] I have searched for similar issues.

joerunde:
## Description

This PR fixes a regression in TP=2 operation.

We used send/recv to implement `all_reduce` ourselves because it's not yet supported. However a recent change has broken both send and recv, so this no longer works. The tests that we previously had would have shown the failing all_reduce, but we had removed them with #215 when we TP=2 was finally fully operational. The only failure now is the full TP=2 test with repeated tokens, but this isn't specific enough to help debug to find the root cause of the problem.

This changes our `all_reduce` op to simply round trip to cpu, which is of course incredibly slow but seems to be the only option for the moment. This also adds tests back for both the all_reduce op and the individual layers that we've overridden for TP=2, so that next time something breaks we might have more info to start with.

Also upgrades torch-spyre to latest main commit, which was unhelpful in resolving the issue with send/recv anyway

## Related Issues

Fixes #276 

## Test Plan

- [ ] unit tests

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ extra-build-variables = { vllm = { VLLM_TARGET_DEVICE = "cpu" } }
 # NB: torch-spyre can only be compiled where `sendnn` is available
 [tool.uv.sources]
 vllm = { git = "https://github.com/vllm-project/vllm", rev = "v0.20.1" }
-torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "4dcfee15c3a9344652f067149ec65c4bf2941890" }
+torch-spyre = { git = "https://github.com/torch-spyre/torch-spyre", rev = "3faf9f1b9c59a280324379cd82e7bdaeb4f3e395" }
 torch = [
   { index = "pytorch-cpu" },
 ]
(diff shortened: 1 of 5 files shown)


joerunde:
@kmehant this works for me on a dev image with these commit hashes:
```
ibm-deeptools:2.0.0-0.main.1+972.4ec0227_0.el10
ibm-flex:2.0.0-0.main.1+307.80bdb14_0.el10
```
(this was a build from June 15)

If you can get the CI updated to that image, then this fix should start passing

kmehant:
@joerunde things seem to be fine now, we can merge it.

joerunde:
Closing- there was a DMA bug that made it into the base images but it's been reverted so send/recv work again.

I'll consider opening a new PR with just the probes added back in, but for now at least CI is green

PR #278, opened 2026-06-17
:bug: Use `cmake --fresh` for vllm

joerunde:
## Description

Builds are currently failing with an error about ninja not being found, which is caused by `cmake` caching the ninja binary location from a previous run's ephemeral build environment which no longer exists. Passing the `--fresh` arg via `CMAKE_ARGS` is supported by vllm to work around this exact problem.

Also, `--refresh` was added in https://github.com/torch-spyre/spyre-inference/pull/227, but this causes packages to always rebuild from source which circumvents some of the uv caching. Pulling it out should improve build speeds

## Checklist

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -105,7 +105,6 @@ jobs:
           uv sync \
             --frozen \
             --verbose \
-            --refresh \
             --group dev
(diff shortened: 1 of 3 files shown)


joerunde:
The one failure is now the strict xpass on gather, but that's already queued in #268 

Merging this to unblock CI

Issue #248, opened 2026-06-09
[Milestone] Model parity with sendnn-inference


joerunde:
## Overview

This issue tracks **Milestone 1** of broad model support for spyre-inference: achieving
parity with the models currently supported by
[sendnn-inference](https://github.com/torch-spyre/sendnn-inference). It is a sub-issue
of the wider model-support epic (#126).

The source of truth for the target model list is
[`sendnn_inference/config/model_configs.yaml`](https://github.com/torch-spyre/sendnn-inference/blob/main/sendnn_inference/config/model_configs.yaml).

---

## Generative Models (decoder-only)

These models use standard dense-transformer architectures already partially covered by
spyre-inference's OOT layer replacements (RMSNorm, QKVParallelLinear, RowParallelLinear,
MergedColumnParallelLinear, RotaryEmbedding, SiluAndMul, ParallelLMHead).

- [x] `ibm-granite/granite-3.3-8b-instruct` — `GraniteForCausalLM`
  _Architecture is exercised end-to-end via the `micro-g3.3-8b-instruct-1b` test model._
- [x] `meta-llama/Llama-3.1-8B-Instruct` — `LlamaForCausalLM`
  _Same OOT layer set as Granite 3.3; needs an end-to-end validation run on Spyre._
- [x] `ibm-granite/granite-4.1-8b` — `GraniteForCausalLM`
  _Granite 4 dense (not MOE); likely compatible, needs E2E test._

---

## Generative Models — Vision-Language (multimodal)

These require a vision encoder / image pre-processing path in addition to the LLM backbone.

- ~[ ] `ibm-granite/granite-vision-3.3-2b` — `LlavaNextForConditionalGeneration`
  _Needs LLaVA-next vision tower (ViT + projection) support on Spyre._~
- [x] `mistralai/Mistral-Small-3.2-24B-Instruct-2506` — `PixtralForConditionalGeneration`
  _Pixtral architecture; needs Pixtral vision encoder support._
- [x] `mistralai/Ministral-3-14B-Instruct-2512-BF16` — `PixtralForConditionalGeneration`
  _Same Pixtral architecture; also needs BF16 dtype support (platform currently enforces
  float16 only)._

---

## Quantized Variants (FP8)

The current platform enforces `torch.float16`; FP8 requires a quantization/dequantization
path compatible with Spyre.

- [x] `ibm-granite/granite-3.3-8b-instruct-FP8` — `GraniteForCausalLM` + FP8 weights
[...]

PR #279, opened 2026-06-17
[Custom Ops] Register SpyreLlama3RotaryEmbedding for Llama-3.1 support

rafvasq:
## Description

To enable meta-llama/Llama-3.1-8B-Instruct 
- Register SpyreLlama3RotaryEmbedding for Llama3RotaryEmbedding
- Add unit tests covering the OOT registration, numerical correctness vs the CPU reference

## Related Issues

Related to #248 

## Checklist


spyre_inference/custom_ops/rotary_embedding.py
```diff
@@ -73,6 +76,58 @@ def forward(
         return out_query, out_key
 
 
+@RotaryEmbeddingBase.register_oot(name="Llama3RotaryEmbedding")
+class SpyreLlama3RotaryEmbedding(Llama3RotaryEmbedding):
```
joerunde: Can this entire class be shrunk down to

```
  @RotaryEmbeddingBase.register_oot(name="Llama3RotaryEmbedding")
  class SpyreLlama3RotaryEmbedding(Llama3RotaryEmbedding, SpyreRotaryEmbedding):
      pass
```

so that we pick up what we need with multiple inheritance instead of copy-pasting .forward from SpyreRotaryEmbedding?

spyre_inference/custom_ops/rotary_embedding.py
```diff
@@ -73,6 +76,58 @@ def forward(
...
+    The frequency scaling (factor, low_freq_factor, high_freq_factor) is computed
+    once during initialization and stored in the cos_sin_cache, so forward() just
+    applies the pre-scaled values.
+    """
+
+    def _apply(self, fn, recurse=True):
+        # Keep cos_sin_cache on CPU so forward() can use it directly.
+        return self
+
+    def forward(
```
joerunde: I think we actually want to override `forward_oot` and not `forward`, because this is a `CustomOp`, not a `PluggableLayer`. (See e.g. rms_norm.py)

I still think we probably want to use multiple inheritance here, but we can fix the override on the base `SpyreRotaryEmbedding`

joerunde (approved):

joerunde:
Force merging- I've seen all the tests pass, we're just getting hung up on runners disconnecting

PR #282, opened 2026-06-18
:arrow_up: support vllm 0.23.0

joerunde:
## Description

Adds support for vllm 0.23.0, with a claude skill distilled from the session for future use.

This is not backwards compatible, it bumps the lower bound

## Test Plan

- [ ] unit tests

## Checklist


joerunde:
bot:next-test

tests/plugin/spyre_testing_plugin/pytest_plugin.py
bohnstingl, line 1: Why does the version bump of `vllm` lead to a change in the KV cache structure? Was this an issue in earlier versions as well, or is this really caused by the version upgrade?
In principle the shape of the KV cache tensor layout is something that we can define in our plugin. Since there is anyways only a single supported attention backend currently, we only need to make sure that our attention backend can handle the KV cache format. @bringlein @jvlunteren any thoughts?

.claude/skills/upgrade-vllm/SKILL.md
```diff
@@ -0,0 +1,221 @@
...
+    # the upstream check falls through and raises. Spyre weights live
+    # on-device, not in a host-side cumem allocator, so a nullcontext is the
+    # correct behaviour.
+    return nullcontext()
+```
+
+Look for analogous wrappers added to `Worker.init_device`, `Worker.compile_or_warm_up_model`, `Worker.execute_model`, etc. The pattern is always: import `nullcontext` (or whatever no-op the CPU branch returns), override the method in `TorchSpyreWorker`, leave a comment that explains why we're not just fixing `is_cpu()`.
+
+If the new wrapper is on `Platform` rather than `Worker`, override on `TorchSpyrePlatform` instead — same principle, same comment.
+
```
bohnstingl, line 158: Will this be handled as an example by Claude? I guess this case will likely not appear in future version bumps, but it might still be good to instruct the AI to carefully check the upstream for such changes.
I was expecting more like a description of "if this fails, then look at..", but I might be wrong.

bohnstingl (commented):
Thank you @joerunde for looking into this! I left some comments that I am not fully clear about.

tests/plugin/spyre_testing_plugin/pytest_plugin.py
joerunde: The kv_cache here is just one created in the test body and passed to us, and the test code changed. I don't think this has any bearing on kv caches created during normal operation.

Here's the relevant bit from the test: https://github.com/vllm-project/vllm/blob/v0.23.0/tests/v1/attention/test_attention_backends.py#L139-L160

Whereas previously we didn't hit any of these if blocks in the test body so the cache given to us to test with wasn't transposed: https://github.com/vllm-project/vllm/blob/v0.20.1/tests/v1/attention/test_attention_backends.py#L477-L498

.claude/skills/upgrade-vllm/SKILL.md
joerunde: True, I think that is what claude was going for with:

> The pattern: `vllm.v1.worker.gpu_worker.Worker.<some method>` adds a new wrapper

I can re-prompt it to reframe things in here better as examples though

rafvasq (commented):
LPGTM 👍

rafvasq (approved):

Issue #280, opened 2026-06-18
Apply list-based indirect access to varlen batch (query) in paged attention kernel


bringlein:
Explore if the "list-based indirect access" can work also for the query tensor of paged attention to avoid slicing on the CPU. 

Follow up from #68, #205, etc.

PR #284, opened 2026-06-19
First attempt to use indirect access for varlen query layout

bringlein:
## Description

vLLM uses the so called `varlen` layout for batching of multiple requests. This format requires no padding, is more memory-efficient and hence the default (without alternative) in upstream vLLM for GPUs. 

In effect it is another indirect access, just like paging of KV Cache: Although the `query` is contiguous in memory, the beginning and end of each single request is determined by the `cu_seqlens` tensor. Therefore, to know which part of the query tensor is the right (partial) request for a given sequence id is an indirect access. 

Consequently, there was the idea to use the same trick/workaround like #205 for this side of paged attention to. 

## Related Issues

Implements #280, follow-up from #205 

## Test Plan

Unit tests and E2E offline inference. 

## Checklist


bringlein:
### Current state

All unit tests and end-to-end inference example pass (locally). However, the list-based access is only happening for decode-only batches with size 1. 
For the other combinations (prefill, or batch size >1) there is silent data corruption happening on Spyre: Either due to `torch.cat`, `torch.stack`, or `torch.unbind`. Which I guess is related to the issue of slicing not supported on Spyre yet. 

CC @bohnstingl @jvlunteren @tdoublep

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl, line 125: Could we also rework this to resolve the warnings that currently appear when running this code? 
For example, by using 
```
sliced_t = output
    for dim, offset in zip(dims, offsets):
        sliced_t = torch.narrow(sliced_t, dim, offset, 1)
    sliced_t.copy_(input)
```

We have to decide whether the `copy_` is the best approach though.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -867,7 +902,8 @@ def _reshape_and_cache(
     @_record_function("spyre_attn::online_softmax")
     def _online_softmax_attention(
         self,
-        query_cpu: torch.Tensor,
+        query_dev: torch.Tensor | None,
+        query_cpu: torch.Tensor | None,
```
bohnstingl, line 906: Given the new indirect access capabilities, do we need to run these operations on CPU?

bringlein: see #399

spyre_inference/v1/attention/backends/spyre_attn.py
bringlein: I guess related to #398 and #400

Issue #53, opened 2026-03-07
Setup GHA test matrix for upstream tests


joerunde:
### Feature description

Use the [Jenkins trigger](https://github.com/marketplace/actions/trigger-jenkins-job) to setup test jobs on our spyre hardware that run subsets of the vllm upstream tests.

### Motivation and context

Our tests are set up to selectively check out and run tests from vllm. (See conftest.py). This is configurable with `UPSTREAM_TESTS_PATHS` and `UPSTREAM_TESTS_PASSING_PATTERNS`, to select for the set of upstream tests to run that we expect to be compatible with our plugin.

We would like to be able to run these tests as a regular part of our CI, and we will clearly need to set up a matrix with multiple test jobs as we onboard more upstream tests- see vllm's massive buildkite pipeline for reference. (Eventually we would like to have public spyre hardware integrated with buildkite, but we're not there yet.)

### Proposed solution

We should be able to set up a test matrix like so in a GHA workflow:

```
jobs:
  trigger-jenkins:
    name: "Jenkins: ${{ matrix.test_config.name }}"
    runs-on: ubuntu-latest
    timeout-minutes: 120
    strategy:
      fail-fast: false
      matrix:
        test_config:
          - name: "Language Generation (facebook models)"
            upstream_tests_paths: "models/language/generation"
            upstream_passing_patterns: "facebook"
            markers: ""
          - name: "Vision (llava_next)"
            upstream_tests_paths: "models/vision"
            upstream_passing_patterns: "llava_next"
            markers: "vision and upstream_passing"
```

then use the `appleboy/jenkins-action@v1` action to trigger the tests on our remote cluster.

❗ All URLs and references to internal job names would need to remain secret❗

### Checklist

- [x] I have searched for similar feature requests

PR #285, opened 2026-06-20
ci: accept the full per-PR dispatch contract (runner_label/ref/repository)

ashokponkumar:
## What
Make `test_each_commit.yaml` accept the **full per-PR ephemeral dispatch contract** that the internal integration pipeline sends, mirroring torch-spyre's `torch_spyre_tests.yaml`:

- declare `workflow_dispatch` inputs `ref`, `repository`, `pr_url_hash`, `build_torch_spyre`, `runner_label` (dropped the earlier `image_label` — `runner_label` is the dispatch's name for it)
- select the per-PR image in all 3 matrix configs via `runs-on: [... , ${{ inputs.runner_label || 'image_spyre_inference' }}]`
- thread `ref`/`repository` into the `build-spyre-inference` composite action so a cross-repo `Test-With:` companion fork/branch is checked out

## Why
The internal pipeline's `workflow_dispatch` always sends `ref`, `repository`, `pr_url_hash`, `build_torch_spyre` and (on the ephemeral path) `runner_label`. GitHub Actions **rejects a dispatch that carries inputs the workflow doesn't declare** ("Unexpected inputs provided" → HTTP 422), so the previous `image_label`-only version would 422 every dispatch from the per-PR ephemeral / downstream-fan-out / Test-With flow. This wires spyre-inference into that flow: a component PR (deeptools/flex) that affects spyre-inference, or an explicit `Test-With: spyre-inference#N`, now lands on a **per-PR runner set** (`image_hf_adapters_pr_<KEY>`) whose image has the PR's RPM **baked in** — no runtime `sudo dnf` (broken on the torch-cicd SCC).

`push` / `pull_request` / `merge_group` runs are **unchanged**: all new inputs default empty / to the standing `image_spyre_inference` set.

## Notes
- DO NOT MERGE casually — coordinated with the spyre-frameworks downstream-fan-out work (internal PR #53). Companion hf-adapters change: torch-spyre/hf-adapters#68.
- Until this merges to `main`, a fan-out dispatch (ref=main) to spyre-inference 422s; dispatching against this branch's ref works (verified).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -81,15 +81,27 @@ jobs:
       matrix:
         include:
           - cfg: Non-distributed spyre tests
-            spyre: spyre_pf_x1__pvc_1tb_x1
+            runs_on:
+              - x86_64
+              - spyre_pf_x1__pvc_1tb_x1
+              - linux
+            image_label: image_spyre_inference
             flags: >-
               -m "not (distributed or upstream)"
           - cfg: Distributed spyre tests
-            spyre: spyre_pf_x2__pvc_1tb_x1
+            runs_on:
+              - x86_64
+              - spyre_pf_x2__pvc_1tb_x1
+              - linux
+            image_label: image_spyre_inference
             flags: >-
               -m distributed
           - cfg: Upstream vLLM tests
-            spyre: spyre_pf_x1__pvc_1tb_x1
+            runs_on:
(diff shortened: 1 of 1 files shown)


ashokponkumar:
@kmehant PTAL

kmehant (approved):

PR #288, opened 2026-06-22
ci: TEMP validate non-root shadow runners (DO NOT MERGE)

ashokponkumar:
Throwaway validation: points test runs-on at image_spyre_inference_shadow to verify the drop-sudo-SA / non-root securityContext change on the shadow runner sets fixes the uv-cache EACCES failure. DO NOT MERGE.

### DIFF
--- a/.github/workflows/test_each_commit.yaml
+++ b/.github/workflows/test_each_commit.yaml
@@ -49,7 +49,7 @@ jobs:
               - x86_64
               - spyre_pf_x1__pvc_1tb_x1
               - linux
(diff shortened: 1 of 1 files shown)


joerunde:
@ashokponkumar the precommit job is failing because it uses the `astral-sh/setup-uv` action to setup uv and python.

I know that we have python and uv pre-installed in these images so we could just remove the `setup-uv` action, but I do think we need to support standard github actions like this on our runners instead of being forced to rely on pre-installed tools.

PR #293, opened 2026-06-24
Keep Q/K/V on Spyre for Attention Computation

jvlunteren:
## Description

Removes the unconditional CPU round-trip for query, key, and value tensors in the attention forward pass.

Adds optional profiling spans using torch.profiler.record_function:
- spyre_attn::forward
- spyre_attn::reshape_and_cache
- spyre_attn::online_softmax

Profiling is controlled via the _ATTN_PROFILING flag (disabled by default)

Forward attention latency (profiling-based) was reduced from ~4 ms to just under 3 ms per decode step (single layer) for `ibm-granite/granite-3.3-8b-instruct`.

## Test Plan

## Checklist

### DIFF
--- a/spyre_inference/v1/attention/backends/spyre_attn.py
+++ b/spyre_inference/v1/attention/backends/spyre_attn.py
@@ -17,9 +17,13 @@
 from dataclasses import dataclass
 from typing import Callable, ClassVar, NamedTuple
 
+import contextlib
+
 import torch
 
 from spyre_inference.custom_ops.utils import convert
+
+_ATTN_PROFILING = False
 
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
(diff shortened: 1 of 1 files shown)


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -670,46 +674,53 @@ def forward(  # ty: ignore[invalid-method-override]
...
-            _target_device,
-        )
+        with torch.profiler.record_function("spyre_attn::forward") if _ATTN_PROFILING else contextlib.nullcontext():
+            k_pages, v_pages = kv_cache
+            # Derive target device from the KV pages — query may arrive on CPU
+            # (e.g. in unit tests) while pages live on the real Spyre device.
+            _target_device = k_pages[0].device
+            num_actual_tokens = attn_metadata.num_actual_tokens
+
+            # Keep Q/K/V on device — slicing, unsqueeze, transpose and contiguous
```
bringlein, line 684: does the slicing produces correct results? in my and @bohnstingl tests it silently (sadly, no error message) created garbage data.

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -201,8 +205,8 @@ def specialized_reshape_and_cache_kernel(
         target_device,
     ):
         for t in range(num_tokens):
-            k_tok = convert(key[t].unsqueeze(1).contiguous(), target_device)
-            v_tok = convert(value[t].unsqueeze(1).contiguous(), target_device)
+            k_tok = key[t].unsqueeze(1).contiguous()
+            v_tok = value[t].unsqueeze(1).contiguous()
```
bringlein, line 209: smart idea!

PR #294, opened 2026-06-25
:bug: Fixes for upstream model tests

joerunde:
# Summary

Gets the Spyre attention path running real model E2E tests (opt-style OPT, llama, qwen, granite) by fixing four boot/correctness issues, adds platform-level KV-cache caps, splits model upstream tests into their own CI bucket, and turns on the actual upstream model tests in the YAML.

## Changes

**Fix opaque-attention dispatcher binding** (`spyre_inference/custom_ops/attention_dispatch.py`, new + `custom_ops/__init__.py` wiring)
- After `dispatch_key="PrivateUse1"` was set, `vllm.LLM(...)` / `vllm serve` failed at engine init: `vllm::unified_attention_with_output` is registered for `PrivateUse1` only, but `SpyreQKVParallelLinear.forward` D2Hs QKV to CPU (downstream `.split()` / scatter — Spyre rejects non-contiguous scatter sources). Dispatcher → CPU → no kernel → boot fails.
- Adds a CPU dispatcher impl that H2Ds q/k/v to Spyre, allocates a fresh Spyre output, re-dispatches the op (now picks the existing PrivateUse1 kernel), and mirrors mutated outputs back. Preserves the torch.compile-friendly opaque boundary; attention output now lands on Spyre, downstream layers don't re-H2D.

**Fix tied-weight lm_head path for OPT-style models** (`spyre_inference/custom_ops/vocab_parallel_embedding.py`)
- Adds `SpyreUnquantizedEmbeddingMethod`. When `config.tie_word_embeddings=True` (e.g. OPT, GPT-2), the model assigns `self.lm_head = self.model.…embed_tokens` and the upstream `LogitsProcessor` calls `lm_head.quant_method.apply(layer, x, bias)`. By that point `_SpyreModelWrapper` has moved hidden_states to CPU while the embedding weight stays on Spyre, so upstream `F.linear` hits torch-spyre's decomposition wrapper with mixed devices.
- The new `apply` converts x to the weight device, runs the linear on Spyre, and returns the result on the original input device — matching `SpyreParallelLMHead.forward_oot`. Wired in via `SpyreVocabParallelEmbedding.__init__`.

**Rewrite attention scatter-write** (`spyre_inference/v1/attention/backends/spyre_attn.py`)
- Spyre's `narrow` / `tensor[i:j] =` / `torch.ops.spyre.overwrite` primitives all silently smear writes across row 0 when the destination is a Spyre tensor scattered along dim=0. The previous per-token `_overwrite(tok, output, [0], [q_start + i])` loop was hitting that.
- `_online_softmax_attention` now builds the full result in a CPU `output_cpu` tensor and copies it back to the Spyre output buffer in a single `output.copy_(...)` at the end. Module docstring on `_overwrite` documents the quirk for future callers.

**Platform-level KV-cache caps** (`spyre_inference/platform.py`)
- New class attrs `MAX_MODEL_LEN_CAP = 128`, `MAX_NUM_SEQS_CAP = 8`. `check_and_update_config` warns and clamps when either is exceeded, before the existing `num_gpu_blocks_override` auto-derivation. Keeps the on-device KV cache inside Spyre's VFIO DMA region for every test (local + upstream) without per-test opt-in.

**Turn on upstream model tests** (`tests/plugin/spyre_testing_plugin/upstream_tests.yaml`)
- `test_common.py::test_models`: `skip` → `mandatory_pass` over gpt2, llama-3.2-1b, qwen2.5-0.5b (opt-125m and tiny-mixtral commented out with reasons). `max_tokens=[2]` for speed, `use_prompt_embeds: [true]` skipped.
- `test_granite.py::test_models`: `skip` → `mandatory_pass` over `micro-g3.3-8b-instruct-1b` at fp16, `max_tokens=[2]`.

**Split model upstream tests into their own CI bucket** (`.github/workflows/test_each_commit.yaml`)
- New `Upstream model vLLM tests` matrix entry: `-m "upstream and model and not distributed"` on a 1-Spyre runner.
- `Upstream vLLM tests` and `Upstream distributed vLLM tests` flags add `and not model` so heavy `vllm.LLM`-based tests no longer pull into kernel/attention buckets.

**Debug skill notes** (`.claude/skills/debug-spyre/attention-notes.md`)
- Tightens the existing notes on Spyre's scatter-write misbehavior to match the rewrite above; future debuggers find the documented pattern instead of rediscovering it.

## Test plan

- [x] `vllm serve facebook/opt-125m --max-model-len 128 --max-num-seqs 4` boots
- [x] `LLM("facebook/opt-125m", ...).generate(...)` returns valid output
- [x] `tests/test_vllm_spyre_next.py::test_basic_model_load` passes (micro-g3.3)
- [x] Cap warn-and-clamp fires on `max_model_len=2048, max_num_seqs=64`; run still succeeds
- [x] `ruff check` + `ruff format --check` clean
- [ ] CI: new `Upstream model vLLM tests` bucket green over gpt2 / llama-3.2-1b / qwen2.5-0.5b / granite


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -766,6 +777,16 @@ def _online_softmax_attention(
...
+        # Spyre's narrow / `tensor[i:j] =` / `torch.ops.spyre.overwrite`
+        # primitives all silently misbehave: an in-place sliced write on a
+        # spyre-resident tensor smears the source across the whole buffer
+        # (writes always hit row 0). Building per-token results into a CPU
+        # tensor and copying the full result back in one shot is the only
+        # pattern that survives. Keep this all-CPU until the final copy_.
+        output_cpu = torch.zeros(
+            num_actual_tokens, num_heads, head_size, dtype=output.dtype, device="cpu"
+        )
+
```
bringlein, line 789: If I understand this comment correctly, that means that the `_overwrite` helper only works for batch size 1?

joerunde: apparently? At least leaving it in is causing a recursion error with granite models at larger batch sizes during the upstream model tests :(

I've started splitting this PR out so that we can get the other fixes in- this one probably needs some more digging

joerunde:
closing in favor of 308 and 313

PR #295, opened 2026-06-25
Add model_impl=transformers support via hf-adapters

### LINKED ISSUE #197: Investigate the HF transformers backend
### Feature description

This issues intends to investigate the HF `transformer` backend for vLLM. It should early surface potential issues and pitfalls for `spyre` when using the model definition not from vLLM directly, but through the HF backend. The ultimate goal is to support a breadth of models through this backend.

### Motivation and context

In vLLM model definitions can be "native", i.e., under `vllm/model_executor/models`, or consumed from HuggingFace transformer through the transformer backend, see https://docs.vllm.ai/en/latest/design/huggingface_integration/. For models that are not yet implemented in vLLM or that are solely implemented on HuggingFace transformers, this option is a viable path to also support the model in vLLM.

### Proposed solution

_No response_

### Checklist

R3hankhan123:
## Description
Upstream transformers backend does not work on spyre because of int64 division not present on torch spyre. This PR encapsulates the HF adaptors repo and transformers backend use HF adaptors model encode rather than Upstream transformers

## Related Issues

Fixes #197 

## Test Plan
Run model with transformers as backend
```
INFO 06-25 07:00:20 [utils.py:233] non-default args: {'dtype': 'float16', 'max_model_len': 1024, 'enable_prefix_caching': True, 'num_gpu_blocks_override': 256, 'model_impl': 'transformers', 'model': 'ibm-granite/granite-3.3-8b-instruct'}
```
## Test Result
```
Processed prompts: 100%|███████████████████████████████████████████████████████████████████████████████████| 1/1 [01:57<00:00, 117.51s/it, est. speed input: 0.07 toks/s, output: 0.17 toks/s]
--------------------------------------------------
Generated text: "\n\nIBM's main businesses include Global Technology Services, Global Business Services, Software, Systems,"
--------------------------------------------------
vllm:kv_cache_usage_perc 0.0
vllm:prefix_cache_queries 8
vllm:prefix_cache_hits 0
vllm:external_prefix_cache_queries 0
vllm:external_prefix_cache_hits 0
vllm:mm_cache_queries 0
vllm:mm_cache_hits 0
vllm:prompt_tokens_cached 0
vllm:cache_config_info 1.0
(EngineCore pid=68999) INFO 06-25 07:09:42 [core.py:1238] Shutdown initiated (timeout=0)
(EngineCore pid=68999) INFO 06-25 07:09:42 [core.py:1261] Shutdown complete
(Worker pid=69133) INFO 06-25 07:09:42 [multiproc_executor.py:775] Parent process exited, terminating worker queues
(Worker pid=69133) INFO 06-25 07:09:42 [multiproc_executor.py:872] WorkerProc shutting down.
```
## Checklist


pyproject.toml
```diff
@@ -32,6 +33,7 @@ spyre_inference = "spyre_inference:register"
 
 [project.entry-points."vllm.general_plugins"]
 spyre_inference_ops = "spyre_inference:register_ops"
+spyre_inference_hf_adaptor = "spyre_inference:register_hf_adapters"
```
tdoublep, line 36: How come we need an additional entry point for this?

tests/test_platform.py
```diff
@@ -110,11 +110,12 @@ def test_block_size_override_non_default_value():
     assert vllm_config.cache_config.block_size == 128
 
 
-def test_block_size_no_override_user_specified():
```
tdoublep: What do these changes have to do with PR?

tests/plugin/spyre_testing_plugin/pytest_plugin.py
```diff
@@ -672,7 +673,7 @@ def _spyre_session_config():
     plugin initialization. Tests that need a different config can still enter
     their own set_current_vllm_config context (nesting is safe).
     """
-    os.environ["VLLM_PLUGINS"] = "spyre_inference,spyre_inference_ops"
+    os.environ["VLLM_PLUGINS"] = "spyre_inference,spyre_inference_ops,spyre_inference_hf_adaptor"
```
tdoublep, line 686: I don't understand why this is needed

spyre_inference/platform.py
```diff
@@ -157,18 +157,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        original_block_size = cache_config.block_size
+        if original_block_size % 64 != 0:
+            new_block_size = ((original_block_size + 63) // 64) * 64
+            logger.warning(
+                "Block size must be a multiple of 64 for the list-based attention "
+                "backend. Overriding block_size from %d to %d.",
+                original_block_size,
+                new_block_size,
+            )
+            cache_config.block_size = new_block_size
```
tdoublep, line 204: Why is this change needed?

spyre_inference/hf_adapters.py
```diff
@@ -0,0 +1,155 @@
...
+        device, dtype = x.device, x.dtype
+        cpu_x = convert(x, "cpu") if device.type != "cpu" else x
+        cpu_pos = (
+            convert(position_ids, "cpu") if position_ids.device.type != "cpu" else position_ids
+        )
+        cos, sin = self.original(cpu_x, cpu_pos)
+        return convert(cos, device, dtype), convert(sin, device, dtype)
+
+
+class HfAdaptersForCausalLM(TransformersForCausalLM):
```
tdoublep, line 117: Since we inherit from TransformersForCausalLM from upstream vLLM, shouldn't the Rope layer already get replaced by vLLM implementation which will then be handled by our custom op? If that isn't happening, then I don't understand why we need to run it on CPU and hf_adapters don't?

spyre_inference/hf_adapters.py
```diff
@@ -0,0 +1,155 @@
...
+    """TransformersForCausalLM wrapper to use HF adapters."""
+
+    def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
+        self._fix_generic_config(vllm_config)
+        super().__init__(vllm_config=vllm_config, prefix=prefix)
+        self._patch_rope()
+        logger.debug("HfAdaptersForCausalLM ready: %s", type(self.model).__name__)
+
+    @staticmethod
+    def _fix_generic_config(vllm_config: VllmConfig) -> None:
```
tdoublep, line 132: Why is this function needed?

tdoublep (commented):
Thanks for making progress on this! In general, it looks quite clean. I have some questions mostly for my understanding.

spyre_inference/platform.py
R3hankhan123: Yeah i was testing if it fixes one error, more details in issue #300

spyre_inference/hf_adapters.py
R3hankhan123: Some models like minsitral have both `params.json` and `config.json` so it was reading from `params.json` and then resolving to `PreTrainedConfig` when using this adaptor as backend instead of `Mistral3ForConditionalGeneration`.

R3hankhan123: So the function is basically forcing vLLM to read config.json also

tests/plugin/spyre_testing_plugin/pytest_plugin.py
R3hankhan123: It was not loading the override and using upstream transformers.

tests/test_platform.py
R3hankhan123: Yeah i was testing some failure related to upstream test and block size. I have now reverted it and opened a separate PR for it.

pyproject.toml
R3hankhan123: My idea was to keep this as a separate entrypoint to make it easier to debug it as a backend. I can merge it into one if thats prefered

spyre_inference/hf_adapters.py
R3hankhan123: From the code i understood this [function](https://github.com/vllm-project/vllm/blob/132765e3560659ff63ebd236203672e991b70e08/vllm/model_executor/models/transformers/base.py#L489)  only replaces RMS norm, linear and conv layers

R3hankhan123: And i will try to run rope the same way ad hf_adapters and will let you know

R3hankhan123: So looks we can reuse rope implementation from hf_adapters 🚀 . I have updated the code

pyproject.toml
```diff
@@ -124,7 +134,8 @@ members = ["tests/plugin"]
 # vllm from source for the cpu backend.
 # NB: This section CANNOT be used to override dependencies to a conflicting version. For example,
 # you cannot add torch==2.10.0 if vllm's build dependencies already specify torch>=2.11.0
-# [tool.uv.extra-build-dependencies]
+[tool.uv.extra-build-dependencies]
+hf-adapters-spyre = ["hatchling"]
```
joerunde: is this just straight up missing from the `hf-adapters-spyre` build?

pyproject.toml
```diff
@@ -86,6 +91,10 @@ override-dependencies = [
     # requests is used by many dependencies, make sure it's patched for CVEs
     "requests>=2.32.4",
 
+    # hf-adapters-spyre requires transformers>=5.10.2, which conflicts with the
+    # testing plugin's pinned 5.5.3. Override to the newer version.
+    "transformers>=5.10.2",
```
joerunde: Let's update the testing plugin to not pin transformers then, ideally it shouldn't get in our way like this

spyre_inference/hf_adapters.py
tdoublep: Thanks. Perhaps once we have vLLM Rope working properly on Spyre, we could try to patch the replacement logic to just have it rewire the HF code to use that. But this is fine for now.

R3hankhan123:
Hi @tdoublep, all the tests have passed except for `populate-rpm-cache`, which isn't related to this PR. Please merge it when it's convenient for you. Thanks!

Issue #18, opened 2026-04-16
[Feature]: Implement EncoderOnlyAttention for bidirectional context modeling


GOavi101:
### Feature description

Implement SpyreEncoderOnlyAttention layer that supports bidirectional attention patterns for encoder-only models (BERT, RoBERTa). This includes full self-attention across all tokens without causal masking, optimized for torch.compile/Inductor backend using the OOT registration pattern.

### Motivation and context

Encoder models require bidirectional attention where each token can attend to all other tokens in the sequence, unlike decoder-only models that use causal masking. Current vllm-spyre-next only supports causal attention patterns. This layer is foundational for BERT-family models used in embedding generation, classification, and retrieval tasks.

### Checklist

- [x] I have searched for similar feature requests

GOavi101:
I tested the following three models, and all of them are working successfully:

- sentence-transformers/all-roberta-large-v1
- ibm-granite/granite-embedding-125m-english
- ibm-granite/granite-embedding-278m-multilingual

tdoublep:
@GOavi101 to create issue to track performance eval .

PR #296, opened 2026-06-25
Add encoder-only attention path for Granite and RoBERTa embeddings on Spyre

### LINKED ISSUE #123: Refactor FMS attention into a vLLM-native attention backend for use via piecewise approach
This issue involves picking up the great work started by @SilverSoldier to create a new attention backend for vLLM that essentially does what the FMS attention module is doing for us today.

The plan is to use this backend via a piecewise compilation approach. Specifically: all layers except attention will be compiled using Inductor (via torch-spyre) and attention layer will be compiled using existing entry-point into deep tools.

This issue has a dependency on torch-spyre/spyre-inference#124

GOavi101:
## Description

Enables vLLM **embedding (pooling) models** — e.g. IBM Granite and RoBERTa-based sentence-transformers — on Spyre through the existing `TorchSpyreModelRunner` and vLLM pooling APIs (`LLM.embed()`). No separate pooling runner.

**In scope (this PR)**

- **CPU integer inputs** for pooling models (`input_ids`, `positions` stay on CPU for bitwise token-type decode and DMA safety).
- **Mixed-device embedding** — CPU `index_select` from a cached CPU mirror of Spyre weight tables; activations returned on Spyre (`mixed_device_embedding`, `nn.Embedding` patch).
- **Encoder-only attention on Spyre** — `_encoder_only_sdpa`: per-sequence bidirectional `F.scaled_dot_product_attention` using dim-0 slices on vLLM's flat varlen layout (no paged KV, no padded `[B,H,L,D]` batch tensors).
- **Pooling `block_size`** — set after `CpuPlatform` defaults, rounded up to a multiple of 64, unless the user passes `--block-size`.
- **Pooling warmup** — single sequence, capped token count for encoder metadata and DMA limits.

**Out of scope (follow-up PR)**

- Pooler still runs on **CPU** (`_SpyreModelWrapper` D2H before `GPUModelRunner._pool`). Moving CLS/MEAN/LAST + L2 normalize to Spyre is planned separately.

## Related Issues
#18 

## Test Plan

Validated with `spyre_inference` from `feature/encoder-pooling-support`, `enforce_eager=True`:

- **Granite 125m** — `ibm-granite/granite-embedding-125m-english`  
  `LLM.embed()` → dim=768, L2 norm ≈ 1.0
-  **RoBERTa-large ST** — `sentence-transformers/all-roberta-large-v1`, `max_model_len=256`  
  dim=1024, L2 norm ≈ 1.0
- **Pooling strategies** — CLS / MEAN / LAST all pass (cosine similarity between strategies ~0.94–0.96 as expected)

## Checklist


spyre_inference/custom_ops/utils.py
```diff
@@ -92,3 +122,50 @@ def convert(tensor, device=None, dtype=None):
         tensor = tensor.to(device=target_device)
 
     return tensor
+
+
+def mixed_device_embedding(
```
bohnstingl: As discussed in https://github.com/GOavi101/spyre-inference/pull/3, the embedding is currently anyways problematic for spyre. Even for non-encoder-only models the encoding falls back to CPU. Since this is a known limitation currently, to me it feels a bit of an overhead to try and split the computation here into spyre and cpu. As an alternative, can we just keep the entire embedding processing on CPU and simplify the logic below? As soon as we have support for embedding from torch-spyre, we can revisit this decision? This also affects the `SpyreVocabParallelEmbedding`

spyre_inference/platform.py
```diff
@@ -146,6 +146,28 @@ def get_device_communicator_cls(cls) -> str:
...
+
+        Returning ``False`` sets ``Attention.use_direct_call = True`` so vLLM
+        invokes ``SpyreAttentionImpl.forward`` directly, letting the impl own
+        device placement for both the encoder (per-sequence SDPA) and decoder
+        (paged KV) paths. Spyre always runs ``enforce_eager=True`` with
+        ``CompilationMode.NONE`` (see ``apply_config_platform_defaults``), so the
+        opaque-op form is not needed to keep attention out of a compiled graph.
+        """
+        return False
+
```
bohnstingl: There is a conflict here. We recently enabled the wrapping of the attention into the opaque layer precisely to avoid that Dynamo can trace into the attention implementation. In theory, both paths should be supported, but at the moment, our `spyre_attn.py` backend cannot be traced fully, so we need the wrapping. The second part of your comment is currently being reworked, i.e., full graph torch.compile for the entire model will be enabled, see https://github.com/torch-spyre/spyre-inference/pull/260.

Let's discuss the precise issue here and find a solution that can also work in both modes.

spyre_inference/platform.py
```diff
@@ -240,11 +262,12 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         # call CpuPlatform.check_and_update_config()
         super().check_and_update_config(vllm_config)
 
+        cache_config = vllm_config.cache_config
+
```
bohnstingl: Was this move necessary by the lint rules?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -670,6 +673,19 @@ def forward(  # ty: ignore[invalid-method-override]
...
+        # Encoder attention: direct Q/K/V, no paged KV cache (vLLM attn_type dispatch).
+        if self.attn_type in (AttentionType.ENCODER_ONLY, AttentionType.ENCODER):
+            return self._forward_encoder_attention(
+                query[:num_actual_tokens],
+                key[:num_actual_tokens],
+                value[:num_actual_tokens],
+                output[:num_actual_tokens],
+                attn_metadata,
+                layer,
+            )
```
bohnstingl: I don't think this is the right place for this dispatching. We should rather carry out this selection in the platform, where we override the CUSTOM attention backend, i.e., here: https://github.com/torch-spyre/spyre-inference/blob/9c476cacc162b02878c2e80e73f545ceb2f50c1d/spyre_inference/platform.py#L73-L75

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,155 @@
...
+# Sequence-length alignment for the encoder attention matmuls.
+#
+# SDPA contracts over ``head_size`` for the score matmul (Q·Kᵀ) and over the
+# (padded) KV length for the value matmul (P·V). Both contraction dims must be a
+# multiple of the Spyre stick size (64 fp16 elements) for the layout solver to
+# assign a valid hardware layout — otherwise compilation fails with
+# ``cannot restickify ... y_var=d2``. ``head_size`` is already a multiple of 64
+# (enforced by ``SpyreAttentionBackend.supports_head_size``); padding the KV
+# length to this granularity satisfies the second matmul and buckets recompiles
+# into tiers instead of one per distinct sequence length.
```
bohnstingl: Do we need to have those detailed comments?

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,155 @@
...
+       — encoder attention is bidirectional.
+    3. Run one ``F.scaled_dot_product_attention`` on ``output.device``.
+    4. Scatter the unpadded per-sequence results back into the flat ``output``.
+
+    For encoder self-attention the KV span equals the query span (no KV cache,
+    no context), so ``seq_lens[s] == q_len`` in practice. The ``min(q_len,
+    seq_lens[s])`` cap is a cheap defensive guard that keeps padding keys out of
+    the softmax should a caller ever pass a right-padded query span.
+
+    Metadata tensors (``query_start_loc``, ``seq_lens``) are read on CPU.
```
bohnstingl: Same here, I think this can be significantly compacted.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,155 @@
...
+    attn_out = F.scaled_dot_product_attention(
+        q_dev, k_dev, v_dev, attn_mask=mask_dev, **sdpa_kwargs
+    )
+
+    # Scatter unpadded results back into the flat output. Pull to CPU first;
+    # Spyre slicing corrupts memory, so per-token writes go through _overwrite.
+    attn_out_cpu = convert(attn_out, "cpu", output.dtype)
+    write_on_spyre = output.device.type == "spyre"
+    if write_on_spyre:
+        from spyre_inference.v1.attention.backends.spyre_attn import _overwrite
```
bohnstingl: I think this should always be fulfilled? The attention backend should be setup such that the `output` is allocated on spyre and is not moved to CPU again.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,155 @@
...
+#
+# SDPA contracts over ``head_size`` for the score matmul (Q·Kᵀ) and over the
+# (padded) KV length for the value matmul (P·V). Both contraction dims must be a
+# multiple of the Spyre stick size (64 fp16 elements) for the layout solver to
+# assign a valid hardware layout — otherwise compilation fails with
+# ``cannot restickify ... y_var=d2``. ``head_size`` is already a multiple of 64
+# (enforced by ``SpyreAttentionBackend.supports_head_size``); padding the KV
+# length to this granularity satisfies the second matmul and buckets recompiles
+# into tiers instead of one per distinct sequence length.
+ENCODER_SEQ_ALIGNMENT = 64
```
bohnstingl, line 39: I think the block_size is already enforced to be a multiple of 64 now? See https://github.com/torch-spyre/spyre-inference/commit/027b7db8e4c1e6117d46b8ce7f6ad16544918181#diff-1a487578e0e1f5f8601711a7448a290078880632e36ef7496ec1d2a8e84d5258

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
bohnstingl: Can we move this import to the top of the file?

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,155 @@
...
+        q_start, q_end = q_starts[s], q_ends[s]
+        q_len = q_end - q_start
+        # [H, L_aligned, D] -> [q_len, H, D] (strided view; materialized below)
+        seq_out = attn_out_cpu[s, :, :q_len, :].transpose(0, 1)
+        if write_on_spyre:
+            for i in range(q_len):
+                tok = convert(seq_out[i : i + 1].contiguous(), target_device.type, output.dtype)
+                _overwrite(tok, output, [0], [q_start + i])
+        else:
+            output[q_start:q_end].copy_(seq_out)
```
bohnstingl: As mentioned above, this path should actually never be invoked

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -116,35 +179,54 @@ def copy_to_cpu(self, n: int | None = None) -> torch.Tensor:
...
-    device — both execute_model (via _model_forward) and _dummy_run
-    (which calls self.model(...) directly).
+        Model outputs (e.g. ``hidden_states``) are moved to CPU after
+        forward. Downstream vLLM paths (pooling, indexing for sampling) expect
+        CPU tensors today. ``SpyreParallelLMHead`` still runs the lm_head
+        matmul on Spyre and handles H2D/D2H for the sampled hidden-state
+        subset in generative models.
+
+    Wrapping at the model level ensures every call site — ``execute_model``,
+    ``_dummy_run``, and warmup — uses the same device rules.
```
bohnstingl: I think this comment is also way too verbose

bohnstingl (changes requested):
I made a pass through the PR. It is great that we are enabling the encoder-only models, thank you for that.

A few points struck me odd though:

1. To me it appears that there is quite some overhead incurred because of the mixed embedding, where some parts are kept on spyre, while others are kept on CPU. Can we simplify this situation? For example, I see two options: a) keep all tensors on spyre initially (as we now also do for the regular embedding) and then just inside the new embedding op transfer all tensors to CPU, carry out the operation on CPU and transfer the results back to spyre or b) keep all relevant tensors for the embedding on CPU in the first place, perform the operation and only transfer the results back to spyre at the end. My preference would be on a).
2. The issue with the opaque attention wrapping is a bit unclear to me. Ideally we should be able to use both approaches, without and with the opaque layer. Currently, we are forced to use the opaque wrapper, because the `spyre_attn.py` is not fully traceable. We should investigate more what the issue here is and figure out a solution. I can help with that for sure.
3. The comments (probably from an AI agent) are really too verbose. I experienced this myself a lot over the past and we have to really insist that the comments are as crips and clean as possible

spyre_inference/platform.py
GOavi101: Agreed — return  False is a workaround. The issue is that SpyreQKVParallelLinear returns Q/K/V on CPU (needed for strided split / paged-KV scatter), but unified_attention_with_output only has a PrivateUse1 kernel, so the opaque path fails on CPU tensors.

GOavi101: we can register a CPU dispatch kernel (same pattern as spyre_convert in #249) that delegates to our attention impl, which already handles device placement internally. That should restore opaque_attention_op=True and work with #260.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
GOavi101: block_size is for the paged decoder KV backend (platform now enforces % 64). Encoder attention doesn't use it — no paged KV. ENCODER_SEQ_ALIGNMENT pads per-batch sequence length for SDPA stick alignment (e.g. len 100 → 128). Same stick size (64), different roles. I can inline 64 and drop the constant if you'd rather avoid the extra name.

spyre_inference/platform.py
```diff
@@ -149,10 +149,36 @@ def get_device_communicator_cls(cls) -> str:
...
+    def opaque_attention_op(cls) -> bool:
+        """Invoke ``SpyreAttentionImpl.forward`` directly, not via the dispatched op.
+
+        ``True`` (CpuPlatform's default) routes attention through
+        ``torch.ops.vllm.unified_attention_with_output``, which dispatches by
+        tensor device. Q/K/V arrive on CPU, so the PrivateUse1-only kernel isn't
+        found. Returning ``False`` sets ``use_direct_call=True`` so the impl owns
+        device placement for both the encoder and decoder paths.
+        """
+        return False
```
maxdebayser: The vllm `Platform.opaque_attention_op` function already returns False. Why is the override needed?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -114,37 +176,36 @@ def copy_to_cpu(self, n: int | None = None) -> torch.Tensor:
...
             if (
                 t is not None
                 and isinstance(t, torch.Tensor)
                 and t.dtype in (torch.int32, torch.int64)
             ):
+                if self._keep_int_inputs_on_cpu:
+                    # Pooling path: normalize to int64 on CPU (no Spyre H2D).
+                    if t.dtype != torch.int64:
+                        return t.to(dtype=torch.int64)
+                    return t
```
maxdebayser: Why is this necessary just for pooling? The non-pooling models also use `torch.nn.Embedding`, or not?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -831,3 +831,54 @@ def _online_softmax_attention(
 
         output.copy_(convert(output_cpu, device=_target_device))
         return output
+
+
+class SpyreEncoderAttentionImpl(SpyreAttentionImpl):
```
maxdebayser: Wouldn't it be better to move this class to `spyre_encoder_attn.py` and just move the code from `forward_encoder_attention` into this class's `forward`?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -59,6 +61,66 @@
...
+# Token count for pooling warmup (single sequence), kept under the DMA limit.
+SPYRE_ENCODER_WARMUP_MAX_TOKENS = 16
+
+
+# Cache of dynamically created CPU-embedding subclasses, keyed by the original
+# embedding class, so ``__class__`` reassignment stays cheap and idempotent.
+_CPU_EMBEDDING_SUBCLASSES: dict[type, type] = {}
+
+
+class _CpuEmbeddingMixin:
```
maxdebayser: Why is this necessary? I thought `SpyreVocabParallelEmbedding` was supposed to take care of this.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,120 @@
...
+    num_kv_heads = key.shape[1]
+    head_size = query.shape[2]
+
+    # Gather host-side (Q/K/V usually arrive on CPU; convert defensively).
+    query_cpu = convert(query, "cpu")
+    key_cpu = convert(key, "cpu")
+    value_cpu = convert(value, "cpu")
+    dtype = query_cpu.dtype
+
+    query_lens = [q_ends[s] - q_starts[s] for s in range(num_seqs)]
```
maxdebayser: is `len(query_start_loc) == num_seqs`? In this case I would expect an out of range error here because q_ends and q_starts have one less element.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,120 @@
...
+) -> torch.Tensor:
+    """Bidirectional self-attention for encoder layers (no KV cache), on Spyre.
+
+    Gathers/pads the flat varlen Q/K/V into one dense batch on CPU, runs a single
+    masked SDPA on ``output.device``, and scatters the unpadded results back. The
+    batched, stick-aligned shape is required for Spyre layout propagation.
+    """
+    qsl = query_start_loc.cpu()
+    q_starts = qsl[:-1].tolist()
+    q_ends = qsl[1:].tolist()
```
maxdebayser: You could just subtract the q_starts from q_ends to obtain the list of lengths so that you don't need to calculate them several times below but the position after the last sequence must be appended so that you don't get a N-1 length tensor. `torch.diff` can also be used for that.

maxdebayser (commented):
I don't understand the embedding layer for pooling models needs special treatment. Isn't it the same logic as for generative models?

spyre_inference/platform.py
GOavi101: the override is needed because we inherit from CpuPlatform, not base Platform.

Base Platform.opaque_attention_op() → False
CpuPlatform.opaque_attention_op() → True (opaque custom op, same as CUDA)
Without our override, Spyre would inherit True from CpuPlatform and route through torch.ops.vllm.unified_attention_with_output. That fails because Q/K/V arrive on CPU (QKV split / paged-KV scatter) while the op kernel is registered only for PrivateUse1.

Returning False is a temporary workaround: direct Python call, no dispatch lookup. The proper fix is registering unified_attention_with_output (and unified_kv_cache_update) with CompositeExplicitAutograd — same pattern as spyre_convert (https://github.com/torch-spyre/spyre-inference/pull/249) — then we can remove this override and inherit True from CpuPlatform again, which is what we want for compile stability and https://github.com/torch-spyre/spyre-inference/pull/260 full-graph compile.

spyre_inference/v1/attention/backends/spyre_attn.py
GOavi101: Yes, agreed — that would be cleaner.

spyre_inference/v1/worker/spyre_model_runner.py
GOavi101: Yes, generative models also embed tokens. The pooling-only flag is not because only pooling has embeddings — it’s because only pooling uses plain nn.Embedding, and we want input_ids / positions to stay on CPU at the wrapper so we never H2D int tensors that Spyre never computes on.

spyre_inference/v1/worker/spyre_model_runner.py
GOavi101: SpyreVocabParallelEmbedding only replaces VocabParallelEmbedding via OOT — not plain nn.Embedding.

Pooling models use both. RoBERTa is the typical case: word_embeddings is VocabParallelEmbedding (OOT handles CPU gather), but position_embeddings and token_type_embeddings are plain nn.Embedding. After model.to("spyre"), stock nn.Embedding.forward would try to gather against Spyre-resident weights with no torch-spyre embedding kernel.

_CpuEmbeddingMixin applies the same CPU-gather → return-on-Spyre pattern to those remaining nn.Embedding modules. The patch intentionally skips VocabParallelEmbedding instances (they are not nn.Embedding subclasses) since the OOT already covers them.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
GOavi101: len(query_start_loc) is num_seqs + 1, not num_seqs . it's the cumulative prefix array (CSR), last entry = total tokens. q_starts = qsl[:-1] and q_ends = qsl[1:] are each length num_seqs, so range(num_seqs) is in bounds. Same convention as the decoder SpyreAttentionImpl.forward (query_start_loc[seq_idx] / [seq_idx + 1]), documented on SpyreAttentionMetadata.query_start_loc.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
GOavi101: Done

spyre_inference/v1/worker/spyre_model_runner.py
maxdebayser: I think I see what you're saying: Roberta models use `nn.Embedding` ([here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/roberta.py#L48)). But I think this is only because of historic reasons. Bert models (which are almos the same), use `VocabParallelEmbedding` ([here](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/bert.py#L49)). 

Can you test with `sentence-transformers/all-MiniLM-L6-v2` for now? It's a Bert model. I'll open an upstream PR to migrate roberta to VocabParallelEmbedding.

maxdebayser: Here is the upstream PR: https://github.com/vllm-project/vllm/pull/47452

maxdebayser: @GOavi101 , the PR was merged, so there will be no more nn.Embedding in granite embedding models. Since the currently pinned version of vllm is still 0.23.0, it will take a couple of weeks for this change to reach your dev environment. So for now, it's better to test with `Bert` models such as `sentence-transformers/all-MiniLM-L6-v2`. Actually, if you look at the implementation, the Roberta models in vllm reuse most of the classes from the Bert implementation.

spyre_inference/v1/worker/spyre_model_runner.py
maxdebayser: This will no longer be necessary as the Roberta models have been migrated to VocabParallelEmbedding in the current vllm `main` branch. See my other comments below.

spyre_inference/v1/worker/spyre_model_runner.py
GOavi101: Sure, I'll test sentence-transformers/all-MiniLM-L6-v2 and use the latest vLLM to test the remaining models

GOavi101:
@maxdebayser 
sentence-transformers/all-MiniLM-L6-v2 working with the changes

```
INFO 07-06 06:20:29 [__init__.py:44] Available plugins for group vllm.platform_plugins:
INFO 07-06 06:20:29 [__init__.py:46] - spyre_inference -> spyre_inference:register
INFO 07-06 06:20:29 [__init__.py:58] Loading plugin spyre_inference
INFO 07-06 06:20:29 [__init__.py:238] Platform plugin spyre_inference is activated
INFO 07-06 06:23:10 [importing.py:45] Triton is installed but 0 active driver(s) found (expected 1). Disabling Triton to prevent runtime errors.
INFO 07-06 06:23:10 [importing.py:69] Triton not installed or not compatible; certain GPU-related functions will not be available.
INFO 07-06 06:23:11 [utils.py:240] non-default args: {'runner': 'pooling', 'dtype': 'float16', 'max_model_len': 128, 'max_num_batched_tokens': 128, 'max_num_seqs': 4, 'disable_log_stats': True, 'enforce_eager': True, 'pooler_config': PoolerConfig(task=None, pooling_type=None, seq_pooling_type='MEAN', tok_pooling_type=None, use_activation=None, dimensions=None, enable_chunked_processing=False, max_embed_len=None, logit_mean=None, logit_sigma=None, logit_bias=None, logit_scale=None, step_tag_id=None, returned_token_ids=None), 'attention_config': AttentionConfig(backend=<AttentionBackendEnum.CUSTOM: None>, flash_attn_version=None, use_prefill_decode_attention=False, flash_attn_max_num_splits_for_cuda_graph=32, tq_max_kv_splits_for_cuda_graph=32, use_cudnn_prefill=False, use_trtllm_ragged_deepseek_prefill=False, use_trtllm_attention=None, disable_flashinfer_prefill=None, disable_flashinfer_q_quantization=False, mla_prefill_backend=None, use_prefill_query_quantization=False, use_fp4_indexer_cache=False, use_non_causal=False), 'model': 'sentence-transformers/all-MiniLM-L6-v2'}
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
INFO 07-06 06:23:16 [config.py:985] Found sentence-transformers tokenize configuration.
INFO 07-06 06:23:17 [model.py:568] Resolved architecture: BertModel
INFO 07-06 06:23:17 [config.py:873] Found sentence-transformers modules configuration.
INFO 07-06 06:23:17 [config.py:900] Found pooling configuration.
INFO 07-06 06:23:19 [model.py:2032] Downcasting torch.float32 to torch.float16.
INFO 07-06 06:23:19 [model.py:1697] Using max model len 128
INFO 07-06 06:23:20 [vllm.py:886] Asynchronous scheduling is disabled.
WARNING 07-06 06:23:20 [vllm.py:942] Enforce eager set, disabling torch.compile and CUDAGraphs. This is equivalent to setting -cc.mode=none -cc.cudagraph_mode=none
WARNING 07-06 06:23:20 [vllm.py:960] Inductor compilation was disabled by user settings, optimizations settings that are only active during inductor compilation will be ignored.
INFO 07-06 06:23:20 [kernel.py:212] Final IR op priority after setting platform defaults: IrOpPriorityConfig(rms_norm=['native'], fused_add_rms_norm=['native'])

    ▄█▀▀█▄  █▀▀▀█▄  █   █  █▀▀▀█▄  █▀▀▀▀    █  █▄   █  █▀▀▀▀ █▀▀▀▀  █▀▀▀█▄ █▀▀▀▀  █▄   █  ▄█▀▀█▄ █▀▀▀▀
    ▀▀▄▄▄   █▄▄▄█▀  ▀▄ ▄▀  █▄▄▄█▀  █▄▄▄     █  █ █  █  █▄▄▄  █▄▄▄   █▄▄▄█▀ █▄▄▄   █ █  █  █      █▄▄▄
         █  █        ▀█▀   █ ▀█▄   █        █  █  █ █  █     █      █ ▀█▄  █      █  █ █  █      █
    ▀▄▄▄█▀  █         █    █   ▀█  █▄▄▄▄    █  █   ▀█  █     █▄▄▄▄  █   ▀█ █▄▄▄▄  █   ▀█  ▀█▄▄█▀ █▄▄▄▄

    version 0.1.dev120    model sentence-transformers/all-MiniLM-L6-v2

WARNING 07-06 06:23:20 [platform.py:224] Block size must be a multiple of 64 for the list-based attention backend. Overriding block_size from 16 to 64.
INFO 07-06 06:23:20 [platform.py:249] Loading worker from: spyre_inference.v1.worker.spyre_worker.TorchSpyreWorker
INFO 07-06 06:23:20 [platform.py:258] Loading scheduler from: vllm.v1.core.sched.scheduler.Scheduler
INFO 07-06 06:23:20 [platform.py:286] Setting num_gpu_blocks_override=4 (4 seqs × 1 blocks/seq)
INFO 07-06 06:23:20 [compilation.py:303] Enabled custom fusions: norm_quant, act_quant
INFO 07-06 06:23:37 [__init__.py:44] Available plugins for group vllm.platform_plugins:
INFO 07-06 06:23:37 [__init__.py:46] - spyre_inference -> spyre_inference:register
INFO 07-06 06:23:37 [__init__.py:58] Loading plugin spyre_inference
INFO 07-06 06:23:37 [__init__.py:238] Platform plugin spyre_inference is activated
INFO 07-06 06:25:03 [importing.py:45] Triton is installed but 0 active driver(s) found (expected 1). Disabling Triton to prevent runtime errors.
INFO 07-06 06:25:03 [importing.py:69] Triton not installed or not compatible; certain GPU-related functions will not be available.
(EngineCore pid=62590) INFO 07-06 06:25:07 [core.py:109] Initializing a V1 LLM engine (v0.21.1.dev0+gad7125a43.d20260615) with config: model='sentence-transformers/all-MiniLM-L6-v2', speculative_config=None, tokenizer='sentence-transformers/all-MiniLM-L6-v2', skip_tokenizer_init=False, tokenizer_mode=auto, revision=None, tokenizer_revision=None, trust_remote_code=False, dtype=torch.float16, max_seq_len=128, download_dir=None, load_format=auto, tensor_parallel_size=1, pipeline_parallel_size=1, data_parallel_size=1, decode_context_parallel_size=1, dcp_comm_backend=ag_rs, disable_custom_all_reduce=True, quantization=None, quantization_config=None, enforce_eager=True, enable_return_routed_experts=False, kv_cache_dtype=auto, device_config=cpu, structured_outputs_config=StructuredOutputsConfig(backend='auto', disable_any_whitespace=False, disable_additional_properties=False, reasoning_parser='', reasoning_parser_plugin='', enable_in_reasoning=False), observability_config=ObservabilityConfig(show_hidden_metrics_for_version=None, otlp_traces_endpoint=None, collect_detailed_traces=None, kv_cache_metrics=False, kv_cache_metrics_sample=0.01, cudagraph_metrics=False, enable_layerwise_nvtx_tracing=False, enable_mfu_metrics=False, enable_mm_processor_stats=False, enable_logging_iteration_details=False), seed=0, served_model_name=sentence-transformers/all-MiniLM-L6-v2, enable_prefix_caching=False, enable_chunked_prefill=False, pooler_config=PoolerConfig(task=None, pooling_type=None, seq_pooling_type='MEAN', tok_pooling_type='ALL', use_activation=True, dimensions=None, enable_chunked_processing=False, max_embed_len=None, logit_mean=None, logit_sigma=None, logit_bias=None, logit_scale=None, step_tag_id=None, returned_token_ids=None), compilation_config={'mode': <CompilationMode.NONE: 0>, 'debug_dump_path': None, 'cache_dir': '', 'compile_cache_save_format': 'binary', 'backend': 'inductor', 'custom_ops': ['all'], 'ir_enable_torch_wrap': False, 'splitting_ops': [], 'compile_mm_encoder': False, 'cudagraph_mm_encoder': False, 'encoder_cudagraph_token_budgets': [], 'encoder_cudagraph_max_vision_items_per_batch': 0, 'encoder_cudagraph_max_frames_per_batch': None, 'compile_sizes': None, 'compile_ranges_endpoints': [128], 'inductor_compile_config': {'enable_auto_functionalized_v2': False, 'size_asserts': False, 'alignment_asserts': False, 'scalar_asserts': False, 'combo_kernels': True, 'benchmark_combo_kernel': True}, 'inductor_passes': {}, 'cudagraph_mode': <CUDAGraphMode.NONE: 0>, 'cudagraph_num_of_warmups': 0, 'cudagraph_capture_sizes': [], 'cudagraph_copy_inputs': False, 'cudagraph_specialize_lora': True, 'use_inductor_graph_partition': False, 'pass_config': {'fuse_norm_quant': True, 'fuse_act_quant': True, 'fuse_attn_quant': False, 'enable_sp': False, 'fuse_gemm_comms': False, 'fuse_allreduce_rms': False, 'fuse_act_padding': False}, 'max_cudagraph_capture_size': None, 'dynamic_shapes_config': {'type': <DynamicShapesType.BACKED: 'backed'>, 'evaluate_guards': False, 'assume_32_bit_indexing': False}, 'local_cache_dir': None, 'fast_moe_cold_start': False, 'static_all_moe_layers': []}, kernel_config=KernelConfig(ir_op_priority=IrOpPriorityConfig(rms_norm=['native'], fused_add_rms_norm=['native']), enable_flashinfer_autotune=False, moe_backend='auto')
(EngineCore pid=62590) INFO 07-06 06:25:07 [multiproc_executor.py:139] DP group leader: node_rank=0, node_rank_within_dp=0, master_addr=127.0.0.1, mq_connect_ip=10.128.18.201 (local), world_size=1, local_world_size=1
INFO 07-06 06:25:15 [__init__.py:44] Available plugins for group vllm.platform_plugins:
INFO 07-06 06:25:15 [__init__.py:46] - spyre_inference -> spyre_inference:register
INFO 07-06 06:25:15 [__init__.py:58] Loading plugin spyre_inference
INFO 07-06 06:25:15 [__init__.py:238] Platform plugin spyre_inference is activated
INFO 07-06 06:26:18 [importing.py:45] Triton is installed but 0 active driver(s) found (expected 1). Disabling Triton to prevent runtime errors.
INFO 07-06 06:26:18 [importing.py:69] Triton not installed or not compatible; certain GPU-related functions will not be available.
[W706 06:26:23.295388260 utils.cpp:41] Warning: numa_migrate_pages failed. errno: 1 (function init_cpu_memory_env)
[W706 06:26:23.295418564 utils.cpp:68] Warning: NUMA binding: Using MEMBIND policy for memory allocation on the NUMA nodes (0). Memory allocations will be strictly bound to these NUMA nodes. (function init_cpu_memory_env)
(Worker pid=62725) INFO 07-06 06:26:24 [__init__.py:33] Registering custom ops for spyre_inference
(Worker pid=62725) INFO 07-06 06:26:24 [rms_norm.py:244] Registered custom op: SpyreRMSNorm
(Worker pid=62725) WARNING 07-06 06:26:24 [cpu_worker.py:105] libtcmalloc is not found in LD_PRELOAD. For best performance, please follow the section `set LD_PRELOAD` in https://docs.vllm.ai/en/latest/getting_started/installation/cpu/ to setup required pre-loaded libraries.
(Worker pid=62725) WARNING 07-06 06:26:24 [cpu_worker.py:105] libiomp is not found in LD_PRELOAD. For best performance, please follow the section `set LD_PRELOAD` in https://docs.vllm.ai/en/latest/getting_started/installation/cpu/ to setup required pre-loaded libraries.
(Worker pid=62725) INFO 07-06 06:26:24 [parallel_state.py:1410] world_size=1 rank=0 local_rank=0 distributed_init_method=tcp://127.0.0.1:34261 backend=cpu:gloo,spyre:spyreccl
(Worker pid=62725) INFO 07-06 06:26:30 [parallel_state.py:1723] rank 0 in world size 1 is assigned as DP rank 0, PP rank 0, PCP rank 0, TP rank 0, EP rank N/A, EPLB rank N/A
(Worker pid=62725) INFO 07-06 06:26:30 [spyre_model_runner.py:233] Loading model sentence-transformers/all-MiniLM-L6-v2...
(Worker pid=62725) Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
(Worker pid=62725) INFO 07-06 06:26:32 [weight_utils.py:663] No model.safetensors.index.json found in remote.
(Worker pid=62725) INFO 07-06 06:26:32 [weight_utils.py:938] Filesystem type for checkpoints: NFS4. Checkpoint size: 0.08 GiB. Available RAM: 984.48 GiB.
(Worker pid=62725) INFO 07-06 06:26:32 [weight_utils.py:900] Prefetching checkpoint files into page cache started (in background, num_threads=8, block_size=16777216 bytes)
Loading safetensors checkpoint shards:   0% Completed | 0/1 [00:00<?, ?it/s]
Loading safetensors checkpoint shards: 100% Completed | 1/1 [00:00<00:00, 368.02it/s]
(Worker pid=62725) 
(Worker pid=62725) INFO 07-06 06:26:32 [default_loader.py:397] Loading weights took 0.02 seconds
(Worker pid=62725) /home/avishek/dt-inductor/torch-spyre/torch_spyre/_monkey_patch.py:107: UserWarning: Backend Spyre does not support int64; downcasting to int32 may change values outside the 32-bit range. You can silence this via warnings.filterwarnings(...) or spyre.set_downcast_warning(False) or TORCH_SPYRE_DOWNCAST_WARN env. variable. (Triggered internally at /home/avishek/dt-inductor/torch-spyre/torch_spyre/csrc/types_mapping.h:101.)
(Worker pid=62725)   return orig_to(self, *args, **kwargs)
(Worker pid=62725) INFO 07-06 06:26:32 [weight_utils.py:872] Prefetching checkpoint files: 10% (1/1)
(Worker pid=62725) INFO 07-06 06:26:32 [weight_utils.py:895] Prefetching checkpoint files into page cache finished in 0.07s
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:268] Spyre-native layer weights moved to spyre
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:269] Model loaded for Spyre in 2.212s.
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:309] Compilation disabled (enforce_eager=True)
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:283] Encoder pooling model: keeping input_ids/positions on CPU
(Worker pid=62725) WARNING 07-06 06:26:32 [spyre_worker.py:118] FIXME: Adding aten::addmm decomposition to work-around torch-spyre crash
(Worker pid=62725) WARNING 07-06 06:26:32 [spyre_worker.py:118] FIXME: Adding aten::addmm.out decomposition to work-around torch-spyre crash
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:335] Warming up model...
(Worker pid=62725) INFO 07-06 06:26:32 [spyre_model_runner.py:348] Pooling warmup: 16 tokens, max_num_seqs=1 (was 4)
(Worker pid=62725) INFO 07-06 06:26:42 [spyre_model_runner.py:358] Warmup done in 10.105s.
(EngineCore pid=62590) INFO 07-06 06:26:42 [core.py:299] init engine (profile, create kv cache, warmup model) took 10.11 s (compilation: 10.11 s)
(EngineCore pid=62590) Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
(EngineCore pid=62590) INFO 07-06 06:26:45 [config.py:985] Found sentence-transformers tokenize configuration.
(EngineCore pid=62590) WARNING 07-06 06:26:45 [scheduler.py:181] Using custom scheduler class vllm.v1.core.sched.scheduler.Scheduler. This scheduler interface is not public and compatibility may not be maintained.
(EngineCore pid=62590) INFO 07-06 06:26:46 [vllm.py:886] Asynchronous scheduling is disabled.
(EngineCore pid=62590) WARNING 07-06 06:26:46 [vllm.py:942] Enforce eager set, disabling torch.compile and CUDAGraphs. This is equivalent to setting -cc.mode=none -cc.cudagraph_mode=none
(EngineCore pid=62590) WARNING 07-06 06:26:46 [vllm.py:960] Inductor compilation was disabled by user settings, optimizations settings that are only active during inductor compilation will be ignored.
(EngineCore pid=62590) INFO 07-06 06:26:46 [kernel.py:212] Final IR op priority after setting platform defaults: IrOpPriorityConfig(rms_norm=['native'], fused_add_rms_norm=['native'])
(EngineCore pid=62590) INFO 07-06 06:26:46 [platform.py:258] Loading scheduler from: vllm.v1.core.sched.scheduler.Scheduler
(EngineCore pid=62590) INFO 07-06 06:26:46 [compilation.py:303] Enabled custom fusions: norm_quant, act_quant
INFO 07-06 06:26:46 [llm.py:394] Supported pooling task: embed
Processed prompts: 100%|████████████████████████████████████████████████| 2/2 [00:07<00:00,  3.86s/it, est. speed input: 0.00 toks/s, output: 0.00 toks/s]
[0] dim=384, L2_norm=1.0000
    head: [-0.03468923270702362, 0.029664630070328712, 0.006180020049214363, 0.026177028194069862, -0.03980189189314842, -0.15991201996803284, 0.06652327626943588, -0.006804426200687885]
[1] dim=384, L2_norm=1.0000
    head: [0.10330843180418015, 0.030526496469974518, 0.02899187058210373, -0.037455152720212936, 0.07914934307336807, -0.055346883833408356, 0.0030686073005199432, -0.008660498075187206]
(EngineCore pid=62590) INFO 07-06 06:26:53 [core.py:1242] Shutdown initiated (timeout=0)
(EngineCore pid=62590) INFO 07-06 06:26:53 [core.py:1265] Shutdown complete
(Worker pid=62725) INFO 07-06 06:26:53 [multiproc_executor.py:775] Parent process exited, terminating worker queues
(Worker pid=62725) INFO 07-06 06:26:53 [multiproc_executor.py:872] WorkerProc shutting down.
```

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -52,6 +58,8 @@ def __init__(self, *args, **kwargs):
             )
 
     def forward(self, input_: torch.Tensor) -> torch.Tensor:
+        weight = cast(torch.Tensor, self.weight.data)
+        weight_cpu = weight.cpu()
```
bohnstingl: What are we trying to do here?

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -61,11 +69,13 @@ def forward(self, input_: torch.Tensor) -> torch.Tensor:
                 self.shard_indices.added_vocab_start_index,
                 self.shard_indices.added_vocab_end_index,
             )
-            masked_input = masked_input.to(input_.device)
         else:
-            masked_input = input_
+            masked_input = input_.cpu()
```
bohnstingl: Can we please do this in a convert(...) statement?

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -61,11 +69,13 @@ def forward(self, input_: torch.Tensor) -> torch.Tensor:
...
 
-        output_parallel = self.quant_method.embedding(self, masked_input.long())
+        output_parallel = convert(
+            F.embedding(masked_input.long(), weight_cpu),
+            device=weight.device,
+        )
```
bohnstingl: Given the discussion that happened [here](https://github.com/torch-spyre/spyre-inference/pull/296#discussion_r3501902756), is this code piece still required, or can we simply reuse the call chain from upstream? 

Or was that discussion unrelated? Then I'd be curious as to why we need this explicit call?

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -61,11 +69,13 @@ def forward(self, input_: torch.Tensor) -> torch.Tensor:
...
                 self.shard_indices.added_vocab_end_index,
             )
-            masked_input = masked_input.to(input_.device)
         else:
-            masked_input = input_
+            masked_input = input_.cpu()
 
-        output_parallel = self.quant_method.embedding(self, masked_input.long())
+        output_parallel = convert(
+            F.embedding(masked_input.long(), weight_cpu),
```
bohnstingl: Can you remind me why we can't simply call F.embedding with the tensors being on spyre and let torch-spyre internally do the fallback? Just like we had it before?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -246,32 +245,25 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
...
 
-        # Keep Attention module buffers (_k_scale, _v_scale, etc.) on CPU.
-        # Attention is nn.Module (not PluggableLayer) so OOT registration is
-        # not possible. Patch _apply to no-op before model.to("spyre") so
-        # the CPU attention backend can access scale buffers without device
-        # mismatch.
+        # Pin Attention buffers (_k_scale, _v_scale, ...) to CPU before
+        # model.to("spyre") via an _apply no-op so the CPU attention backend
+        # reads them without a device mismatch (these are nn.Module, not
+        # PluggableLayer, so OOT registration isn't possible).
```
bohnstingl: Is this change necessary?

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -246,32 +245,25 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
...
                 module._apply = lambda fn, recurse=True, _m=module: _m
 
-        # Move layer weights to Spyre device.
-        # SpyreCpuFallbackMixin._apply() no-op keeps CPU fallback layer
-        # weights on CPU (linear, embedding, rotary).
-        # Spyre-native layers (RMSNorm, SiluAndMul, ParallelLMHead) get
-        # their weights moved.
+        # Move all weights to Spyre (embeddings included; the gather still runs
+        # on CPU inside SpyreVocabParallelEmbedding — torch-spyre has no
+        # embedding kernel yet).
```
bohnstingl: Is this level of detail necessary? Can we just say "# Move all weights to Spyre "? I think the explicit comment about the embedding might become outdated very quickly.

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -114,31 +120,24 @@ def copy_to_cpu(self, n: int | None = None) -> torch.Tensor:
...
-
-    Wrapping at the model level ensures ALL call sites get the right
-    device — both execute_model (via _model_forward) and _dummy_run
-    (which calls self.model(...) directly).
+    """Converts model inputs/outputs at the device boundary, for every call site.
+
+    Integer inputs move to Spyre int64; outputs move to CPU. Pooling and
+    generative models share this path: embeddings run through
+    ``SpyreVocabParallelEmbedding``, which gathers on CPU internally, so no
+    special-casing of integer inputs is needed here.
```
bohnstingl: Why this change of the docstrings? Was the old one incorrect?
I'd either leave it as it was or simplify it fully and just have "Converts model inputs/outputs at the device boundary for the model"

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -246,32 +245,25 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
...
 
         # Compile for Spyre (no-op if enforce_eager=True)
         self._compile_for_spyre()
 
-        # Wrap model so ALL forward() calls to the entire model,
-        # for example in execute_model, _dummy_run, etc.,
-        # automatically convert Spyre outputs to CPU. This ensures downstream
-        # indexing (logits_indices), lm_head (CPU weights), and sampling all
-        # receive CPU tensors without needing per-call-site overrides.
+        # Device boundary: integer inputs move to Spyre int64, outputs to CPU.
```
bohnstingl: I'd remove this comment tbh.

spyre_inference/platform.py
bohnstingl: Do you actually see an error with this? I checked on main and the op `torch.ops.vllm.unified_attention_with_output` is registered under the PrivateUse1 DispatchKey, but the attention is successfully called with q and k residing on `spyre`, while v is on `cpu`.

Moreover, https://github.com/bohnstingl/spyre-inference/pull/2 will remove this problem altogether, because the instead of performing the split operation, the weight matrix is unfused and then all the inputs should reside on spyre. 

I am a little cautious with enforcing the direct call to attention, as this might trigger all sorts of other problems with the attention implementation being traced now in compile mode.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
```diff
@@ -0,0 +1,156 @@
...
+
+        num_heads = query.shape[1]
+        num_kv_heads = key.shape[1]
+        head_size = query.shape[2]
+
+        # Gather host-side (Q/K/V usually arrive on CPU; convert defensively).
+        query_cpu = convert(query, "cpu")
+        key_cpu = convert(key, "cpu")
+        value_cpu = convert(value, "cpu")
+        dtype = query_cpu.dtype
```
maxdebayser, line 103: I see that a lot of things are actually done on CPU here. Is it because it doesn't work on torch-spyre yet? Maybe add a comment explaining this, because ideally we want to do as much as possible on the device to prevent synchronization between cpu and the device.

spyre_inference/custom_ops/vocab_parallel_embedding.py
GOavi101: this is the CPU embedding gather. torch-spyre has no embedding kernel, so the table stays on Spyre with the rest of the weights, but the actual F.embedding lookup has to run on CPU. To do that we need a CPU copy of the table:

- self.weight.data is the embedding table, which lives on Spyre (all model weights are moved there in load_model).

- we copy it to CPU so F.embedding(indices_cpu, weight_cpu) can run, then convert() the result back to the table's device for the downstream layers.

spyre_inference/v1/attention/backends/spyre_encoder_attn.py
GOavi101: added comments

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -269,24 +270,20 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
...
-        # for example in execute_model, _dummy_run, etc.,
-        # automatically convert Spyre outputs to CPU. This ensures downstream
-        # indexing (logits_indices), lm_head (CPU weights), and sampling all
-        # receive CPU tensors without needing per-call-site overrides.
-        self.model = _SpyreModelWrapper(self.model, self._spyre_device)
+        is_pooling = self.model_config.runner_type == "pooling"
+        self.model = _SpyreModelWrapper(
+            self.model,
+            self._spyre_device,
+            keep_int_inputs_on_cpu=is_pooling,
```
maxdebayser: Wait, why was this changed now? What is the difference between pooling models and generative models w.r.t. inputs?

spyre_inference/custom_ops/vocab_parallel_embedding.py
```diff
@@ -42,23 +43,27 @@ def __init__(self, *args, **kwargs):
...
                 self.shard_indices.added_vocab_start_index,
                 self.shard_indices.added_vocab_end_index,
             )
-            masked_input = convert(masked_input, device=input_.device)
         else:
-            masked_input = input_
+            masked_input = input_cpu
 
-        output_parallel = self.quant_method.embedding(self, masked_input.long())
+        output_cpu = F.embedding(masked_input.long(), weight_cpu)
```
maxdebayser: I don't think that's right. What if a quantized model is loaded? Isn't the right way to plug in a quant_method?

maxdebayser:
Yesterday this PR was in a good shape but now it has regressed a bit. Usually pooling models are pure encoder models, but there are exceptions like Qwen3-Embedding. The difference between encoder and decoder is the causal mask and the fact that that no tokens are generated, so the complexity of paged attention is not required and you can do a plain torch sdpa. Most bert models also don't use rotary embeddings.

Considering the above, there shouldn't be any differences with decoder models regarding the movement of inputs to and from CPU. Also, testing if a model is a pooling model is not always valid because there are decoder pooling models,

spyre_inference/custom_ops/vocab_parallel_embedding.py
GOavi101: Good point — in general yes, quant_method is the right plug-in. Here we already guard in __init__ and raise NotImplementedError for anything other than UnquantizedEmbeddingMethod, so quantized embeddings aren't supported on Spyre yet.

For the unquantized case, quant_method.embedding is just F.embedding(input_, layer.weight). We call F.embedding directly only to control device placement: CPU indices + CPU weight, then move the result back to Spyre. quant_method.embedding always uses layer.weight on Spyre, which hits a device mismatch when pooling inputs stay on CPU.

spyre_inference/v1/worker/spyre_model_runner.py
GOavi101: Generative models still move input_ids/positions to Spyre. Pooling/encoder models keep them on CPU because 
(a) small int H2D hits a DmaWorker: bad_function_call during warmup
(b) the BERT/RoBERTa path needs CPU indices for token_type decode + mixed-device embedding gather anyway.

maxdebayser: a) Ok, so I think that perhaps needs some investigation: why do these pooling model hit the Dma worker error while the others don't
b) The `token_type` embeddings work in the same exact way as the position and token embeddings. Why do they need special treatment on CPU?

GOavi101: (b) you're right. token_type_embeddings works exactly like word/position embeddings. The only encoder-specific bit is _decode_token_type_ids (bitwise unpack on input_ids), and that doesn't require the inputs to be on CPU — SpyreVocabParallelEmbedding already gathers on CPU internally regardless of where indices arrive. So there's no real justification for special-casing token_type on CPU. I overstated that.

spyre_inference/platform.py
GOavi101: I'll remove this and test it once https://github.com/torch-spyre/spyre-inference/pull/338 is merged.

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -235,15 +226,48 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
+                "not found; Spyre encoder patch needs updating for this "
+                "vLLM version."
+            )
+
+        if model_config.runner_type != "pooling":
+            return
+
+        # Only BertModel is verified on Spyre.
+        if "BertModel" not in model_config.architectures:
+            return
```
maxdebayser: I would leave this out. Because once the VocabParallelEmbeddings PR arrives in the vLLM version pinned by sendnn-inference, support for RobertaModels won't require any code changes.

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -344,7 +356,22 @@ def warming_up_model(self) -> None:
...
+                num_tokens = min(num_tokens, SPYRE_ENCODER_WARMUP_MAX_TOKENS)
+                saved_max_num_seqs = self.scheduler_config.max_num_seqs
+                try:
+                    self.scheduler_config.max_num_seqs = 1
+                    logger.info(
+                        "Pooling warmup: %d tokens, max_num_seqs=1 (was %d)",
+                        num_tokens,
+                        saved_max_num_seqs,
+                    )
+                    self._dummy_run(num_tokens)
```
maxdebayser, line 499: I thought that the idea was to do this only for eager mode, but I don't see a check for that mode here.

spyre_inference/custom_ops/vocab_parallel_embedding.py
maxdebayser: Maybe add a TODO or FIXME comment here to revert to quant_method once the issues that are preventing its use are fixed.

maxdebayser:
@bohnstingl , from the point of view of embedding models logic this is looking good already. I've just left a few minor recommendations. So now I guess the question is whether the `VocabParallelEmbedding` changes will work with the other models.

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -175,22 +181,9 @@ def copy_to_cpu(self, n: int | None = None) -> torch.Tensor:
...
-
-    Output conversion (Spyre → CPU):
-        The model's final hidden_states come out on Spyre. Downstream
-        operations (indexing via logits_indices, sampling) run on CPU.
-        The lm_head matmul runs on Spyre via SpyreParallelLMHead,
-        which handles H2D/D2H for the sample_hidden_states subset.
-
-    Wrapping at the model level ensures ALL call sites get the right
-    device — both execute_model (via _model_forward) and _dummy_run
-    (which calls self.model(...) directly).
```
bohnstingl, line 203: Why do we have to remove those comments? They are still up-to-date, no? I.e., the data transfer still exists

spyre_inference/v1/worker/spyre_model_runner.py
```diff
@@ -344,11 +364,6 @@ def load_model(self, load_dummy_weights: bool = False) -> None:
         # Compile for Spyre (no-op if enforce_eager=True)
         self._compile_for_spyre()
 
-        # Wrap model so ALL forward() calls to the entire model,
-        # for example in execute_model, _dummy_run, etc.,
-        # automatically convert Spyre outputs to CPU. This ensures downstream
-        # indexing (logits_indices), lm_head (CPU weights), and sampling all
-        # receive CPU tensors without needing per-call-site overrides.
```
bohnstingl, line 393: Why remove?

spyre_inference/platform.py
```diff
@@ -190,8 +190,22 @@ def get_device_communicator_cls(cls) -> str:
...
+        # Encoder (pooling) layers have no KV cache and run bidirectional SDPA;
+        # decoders use the paged backend. vLLM passes attn_type via the selector
+        # config, so the choice lives here rather than as a branch in the impl.
+        from vllm.v1.attention.backend import AttentionType
+
+        attn_selector_config = kwargs.get("attn_selector_config") or (args[0] if args else None)
+        attn_type = getattr(attn_selector_config, "attn_type", None)
+        if attn_type in (AttentionType.ENCODER, AttentionType.ENCODER_ONLY):
+            return cls._encoder_backend_path
         return AttentionBackendEnum.CUSTOM.get_path()
```
bohnstingl: Could we simplify and consolidate this code piece a bit. For example, into something like

```
@classmethod
def get_attn_backend_cls(cls, selected_backend, *args, **kwargs) -> str:
    # Encoder (pooling) layers have no KV cache and run bidirectional SDPA;
    # decoders use the paged backend. vLLM passes attn_type via the selector
    # config, so the choice lives here rather than as a branch in the impl.
    from vllm.v1.attention.backend import AttentionType

    attn_selector_config = kwargs.get("attn_selector_config") or (args[0] if args else None)
    attn_type = getattr(attn_selector_config, "attn_type", None)
    if attn_type in (AttentionType.ENCODER, AttentionType.ENCODER_ONLY):
        # Specific spyre attention for encoder models
        _backend_path = (
            "spyre_inference.v1.attention.backends.spyre_encoder_attn.SpyreEncoderAttentionBackend"
        )
    else:
        # Standard spyre attention
        _backend_path = "spyre_inference.v1.attention.backends.spyre_attn.SpyreAttentionBackend"
    
    # Register the specific spyre attention implementation as the CUSTOM backend.
    register_backend(AttentionBackendEnum.CUSTOM, _backend_path)
    
    return AttentionBackendEnum.CUSTOM.get_path()
```

PR #298, opened 2026-06-25
Add profiling annotations to attention computation on Spyre

jvlunteren:
## Description

Adds optional profiling spans using torch.profiler.record_function:
- spyre_attn::forward
- spyre_attn::reshape_and_cache
- spyre_attn::online_softmax

Profiling is controlled via the _ATTN_PROFILING flag (disabled by default)

## Test Plan

## Checklist


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -670,39 +675,44 @@ def forward(  # ty: ignore[invalid-method-override]
...
-            v_pages,
-            attn_metadata.slot_block_indices[:num_actual_tokens],
-            attn_metadata.slot_block_offsets[:num_actual_tokens],
-            _target_device,
-        )
+        with (
+            torch.profiler.record_function("spyre_attn::forward")
+            if _ATTN_PROFILING
+            else contextlib.nullcontext()
+        ):
```
bringlein: that looks very useful, thanks! 
But can we maybe implement this as a decorator? so to decorate the functions we want to annotate? 

I don't know what a good trade-off here is: to have fine-grained annotations and a lot of code changes for them (yes, only indents) or to stick to the function scoping with very little change for profiling? Does it make sense?

bohnstingl: I am somewhat in favor of the decorator approach. However, I think the downside with this would be that we can only decorate functions and not specific code segments?
To me though there is a bigger question: Should the profiling statements actually really reside in the main path, or just locally or in some features branches?

jvlunteren: I replaced these with decorators.

PR #299, opened 2026-06-25
Add softmax switch: numerically stable (CPU fallback) vs Spyre-only execution

jvlunteren:
## Description

This PR introduces a `_NUM_STABLE_SOFTMAX` flag in `spyre_attn.py` to toggle between two online softmax implementations. By default (True), it uses a numerically stable approach based on `scores.max()`, which triggers a CPU fallback via `aten::argmax`. When disabled (False), it switches to a straightforward `torch.exp(scores)` computation, avoiding the CPU round-trip and extra rescaling operations, but sacrificing numerical stability. This option is intended solely for testing and profiling, until `aten::argmax` is supported on Spyre.

## Test Plan

Running an experiment with the `ibm-granite/granite-3.3-8b-instruct` model using the prompt "_What do you know about Zurich?_" produces the same output for both settings of `_NUM_STABLE_SOFTMAX`:
"_\n\nZurich is the largest city in Switzerland and the capital of the canton of Z_"
(with `max_tokens=40` and `temperature=0.0`).

## Checklist


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -39,6 +39,12 @@
 
 logger = init_logger(__name__)
 
+# When False, uses plain softmax without per-tile max shift: eliminates the aten::argmax
+# CPU fallback and the rescaling ops on every page beyond the first. Numerically unstable
+# for large score magnitudes. When True, uses the numerically stable online softmax with
+# per-tile max (requires a CPU round-trip via aten::argmax on every page).
```
bohnstingl, line 45: Is this fallback to CPU for torch.max confirmed?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -249,32 +255,44 @@ def specialized_paged_attn_kernel(q, k_pages, v_pages, page_indices, mask_tiles,
...
+                        tile_probs, None, v_pages, page_idx, transform_b=lambda t: t.unsqueeze(1)
+                    )
+                    tile_sum = tile_probs.sum(dim=-1, keepdim=True)
+                else:
+                    assert tile_sum is not None
+                    assert tile_output is not None
+                    tile_output += _indirect_matmul_mock(
+                        tile_probs, None, v_pages, page_idx, transform_b=lambda t: t.unsqueeze(1)
+                    )
+                    tile_sum = tile_sum + tile_probs.sum(dim=-1, keepdim=True)
```
bohnstingl, line 300: If the maximum does not trigger an explicit fallback to CPU, I feel like there is some duplicate code that could be simplified. In particular, when we unset `_NUM_STABLE_SOFTMAX`, can we then just enforce the `scores_max=0` and get the same behavior without a separate code path? For example, what about this code below

```
if _NUM_STABLE_SOFTMAX:
    scores_max = scores.max(dim=-1, keepdim=True)[0]
else:
    scores_max = 0.

if i == 0:
    tile_max = scores_max
    tile_probs = torch.exp(scores - tile_max)
    tile_output = _indirect_matmul_mock(tile_probs, None, v_pages, page_idx, transform_b=lambda t: t.unsqueeze(1))
    tile_sum = tile_probs.sum(dim=-1, keepdim=True)
else:
    new_max = torch.maximum(tile_max, scores_max)
    rescale = torch.exp(tile_max - new_max)
    tile_output = tile_output * rescale
    tile_sum = tile_sum * rescale
    tile_probs = torch.exp(scores - new_max)
    tile_output += _indirect_matmul_mock(tile_probs, None, v_pages, page_idx, transform_b=lambda t: t.unsqueeze(1))
    tile_sum = tile_sum + tile_probs.sum(dim=-1, keepdim=True)
    tile_max = new_max
```

bohnstingl (commented):
My comments hinge on whether torch.max triggers a CPU roundtrip or not

jvlunteren:
This PR is not needed anymore now that `torch.amax` is supported.

PR #303, opened 2026-06-25
Add flag to force torch.compile on Spyre attention

jvlunteren:
## Description

This PR adds a module-level `_FORCE_COMPILE_ATTN` flag in `spyre_attn.py` (default: `False`) that forces `torch.compile(dynamic=False)` to be applied to the paged attention kernel, bypassing the vLLM compilation configuration. On Spyre, where `CompilationMode.NONE` is set, the standard compilation path is effectively a no-op. This flag enables evaluation of the compiled attention path without altering vLLM settings.

This flag is intended for experimental use. The default value (False) ensures that existing behavior remains unchanged.

## Test Plan

## Checklist


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -39,6 +39,10 @@
 
 logger = init_logger(__name__)
 
+# When True, wraps _create_compilable_page_attn with torch.compile(dynamic=False) directly.
+# When False (default), uses _maybe_compile which is a no-op on Spyre (CompilationMode.NONE).
+_FORCE_COMPILE_ATTN = False
```
bringlein: I think that is a helpful idea, but can we overwrite this flag with an env variable?

bohnstingl: I would agree to this here. In upstream vLLM (as well as in our plugin at the moment), the attention is wrapped in an opaque layer and torch.compile doesn't trace inside it. To me there are two options that we can choose from

1. Follow this approach with "forcing torch.compile". We might want to use though an environmental variable and simplify the _maybe_compile function. We don't need any of the additional check, but simply trigger compilation via. torch.compile(fn) in case the environmental variable is set
2. Make our attention implementation traceable and remove the opaque boundary. This would be the "cleaner" long-term solution, but it is more unrealistic to be honest, because this would require that the entire attention implementation can be compiled into a big graph, which (given the current spyre limitations) might not be possible to do.

So to me, 1) would be a good option to go

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -645,8 +649,9 @@ def _get_reshape_fn(self, num_tokens: int):
     def _get_attn_fn(self, num_blocks: int, padded_query_len: int):
         key = (num_blocks, padded_query_len)
         if key not in self._attn_fns:
-            self._attn_fns[key] = _maybe_compile(
-                _create_compilable_page_attn(num_blocks, padded_query_len)
+            fn = _create_compilable_page_attn(num_blocks, padded_query_len)
+            self._attn_fns[key] = (
+                torch.compile(fn, dynamic=False) if _FORCE_COMPILE_ATTN else _maybe_compile(fn)
```
tdoublep: Perhaps it would it better inside the `_maybe_compile` function itself?

spyre_inference/v1/attention/backends/spyre_attn.py
jvlunteren: Done.

spyre_inference/v1/attention/backends/spyre_attn.py
jvlunteren: Done: I created a separate` _maybe_compile` function for online softmax because cache-and-reshape currently fails when being compiled.

PR #308, opened 2026-06-25
⚡ Limit batch volume to allow models to boot

joerunde:
## Description

Extracts just the upstream model test enablement and the one fix for the `RAS::VFIO::MapDMAFailed` error from #294 to start unblocking some of the models. 

## Related Issues

Related to #248, as is these models will never boot without overriding to lower batch sizes.

Also adds a test for llama which was recently enabled (cc @rafvasq)

## Test Plan

Upstream model tests

## Checklist


tests/plugin/spyre_testing_plugin/upstream_tests.yaml
```diff
@@ -59,20 +59,26 @@ tests:
...
           params:
-            override:
+            allow:
               model:
-                - facebook/opt-125m
-                - ibm-granite/granite-3.3-2b-instruct
-                - Qwen/Qwen3-0.6B
+                # - openai-community/gpt2 # needs opaque-attention CPU dispatch + tied-weight lm_head fix
+                # - facebook/opt-125m # requires a torch-spyre fix for int adds
+                - meta-llama/Llama-3.2-1B-Instruct
```
rafvasq, line 74: FYI, I tried using unsloth/Llama-3.2-1B as an alternative to avoid gating, but it failed with:

```
FAILED ::test_models[False-False-5-32-unsloth/Llama-3.2-1B] - ValueError:
    No example model defined for unsloth/Llama-3.2-1B; please update this
  file.
```

maxdebayser (approved):
LGTM

joerunde:
Just added llama 3.2 1b to the cache so these tests should be able to run now

tests/plugin/spyre_testing_plugin/upstream_tests.yaml
joerunde: yeah- I thought that would be a nice workaround for needing a gated model, but it turns out the models available for this test are hardcoded

PR #310, opened 2026-06-25
CustomOp cleanup

bohnstingl:
## Description

This PR does a general cleanup of the CustomOps that are currently available in spyre-inference. In particular, it 

- Removes unnecessary churn in the documentations
- Cleans up dead or unnecessary OOT implementations of CustomOps
- Rework and simplify CustomOps where possible

## Related Issues

NA

## Test Plan

Changes are transparent to the user, so no new tests are required. However, the existing tests had to be changed.

## Checklist


spyre_inference/custom_ops/parallel_lm_head.py
joerunde: I think we can drop this convert?

The new [wrapping in the model runner](https://github.com/torch-spyre/spyre-inference/pull/310/changes#diff-d68d9224c2242aa7c9e8ce1af15e197fb181112e341cc86eab0b0e05242fef5aR174-R186) will immediately bounce this back to cpu, so this amounts to `out_cpu_no_pad.to("spyre").to("cpu")` when the input device is spyre

spyre_inference/custom_ops/rms_norm.py
```diff
@@ -12,233 +12,78 @@
...
-Limitations:
-    Currently the implementation in `forward_spyre` is similar to the
-    upstream implementation in `forward_static` from vllm/model_executor/layers/layernorm.py,
-    but it DOES NOT use the promotion of the data types, as this is not
-    yet supported in torch-spyre.
+"""Spyre OOT replacement for RMSNorm.
+
+Spyre constraints:
+    - No dtype promotion to float32 (not yet supported in torch-spyre)
+    - variance_epsilon is materialized as a tensor (scalar broadcast unsupported).
```
joerunde: The diff looks like it's removing the line that does this:
```
        variance_epsilon_t = torch.full(
            x.shape, variance_epsilon, dtype=torch.float16, device=x.device
        )
```

is this still correct or are we using scalar broadcasting now?

bohnstingl: Yes, this is intentional. The "tensor" form was required for earlier versions of `torch-spyre`, but meanwhile the capabilities have improved and we can do the broadcasting that upstream also does. At least in my E2E tests and in the CI tests, it works without issues.

spyre_inference/custom_ops/parallel_lm_head.py
bohnstingl: True, good catch. I guess this is a question about consistency in the layers. In general the issue is that we cannot do the final sampling, that happens in the runner after computing the lm_head, on spyre because of slicing. That's why we have to bounce back in the runner function that you are referring to.

On the other hand, the current convention for all our CustomOps is that the inputs arrive on spyre and the outputs are also placed on spyre. So these two aspects clash here and hence the current solution. Do you have a strong preference towards on way or another?

spyre_inference/custom_ops/rms_norm.py
joerunde: Right, so should we update the comment then? The `(scalar broadcast unsupported)` bit will probably confuse poor claude

spyre_inference/custom_ops/parallel_lm_head.py
joerunde: Gotcha- I guess I don't have a strong preference then. Since we're not currently trying to optimize for performance it's probably fine to leave the extra hop in so that we honor the custom op contract of leaving the outputs on the same device as the inputs.

joerunde (approved):
lpgtm!

spyre_inference/custom_ops/__init__.py
```diff
@@ -32,5 +32,4 @@
 def register_all():
     logger.info("Registering custom ops for spyre_inference")
     rotary_embedding.register()
-    rms_norm.register()
```
tdoublep, line 35: Why don't we need to register rms_norm but we need to register rotary_embedding?

spyre_inference/custom_ops/rotary_embedding.py
```diff
@@ -12,82 +12,135 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Spyre OOT replacement for RotaryEmbedding (CPU fallback).
+"""Spyre OOT replacement for RotaryEmbedding.
 
-Remove this file once Spyre natively supports rotary embedding ops.
+The CPU body is wrapped in a `direct_register_custom_op` so torch.compile does
+NOT trace into RotaryEmbedding.forward_native.
```
tdoublep, line 18: What is the problem with letting compile trace into it?

spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -84,7 +84,12 @@ def _overwrite(
     if output.device.type == "spyre":
         # `torch.ops.spyre.overwrite` is dynamically registered, so its
         # signature is opaque to the type checker (ParamSpec resolves to `...`).
-        torch.ops.spyre.overwrite(input, output, dims, offsets)  # ty: ignore[invalid-argument-type]
+        torch.ops.spyre.overwrite(
+            input,  # ty: ignore[invalid-argument-type]
+            output,  # ty: ignore[invalid-argument-type]
+            dims,  # ty: ignore[invalid-argument-type]
+            offsets,  # ty: ignore[invalid-argument-type]
+        )
```
tdoublep, line 92: why this change?

spyre_inference/custom_ops/rms_norm.py
bohnstingl: I removed the comment to not cause confusion. At this point, it is the standard handling of upstream vLLM.

spyre_inference/custom_ops/__init__.py
bohnstingl: The implementation of `rms_norm` is very close to the upstream vLLM implementation. It has all tensors on Spyre and does perform every operation on Spyre. Therefore, we don't need to wrap anything behind an opaque layer, hence we also don't need to register anything (doing the `direct_register_custom_op`), a simple import is enough.

On the other hand, in RoPE not all tensors are kept on Spyre, for example the `cos_sin_cache`. Also, not all operations can be performed on Spyre. Therefore, we need to wrap the operation in an opaque layer for torch.compile.

spyre_inference/custom_ops/rotary_embedding.py
bohnstingl: There are issues with torch-spyre, because the input to the layer, e.g., `positions` are on spyre, while the `cos_sin_cache` cache is cpu. So this would correspond to a graph with mixed inputs, some on spyre and some on cpu. I am hoping that with the rework of RoPE, will can get rid of all of those issues

tdoublep: >So this would correspond to a graph with mixed inputs, some on spyre and some on cpu.

Is this inherently a problem?

bohnstingl: As of today in `torch-spyre`, yes. This is not possible. We can discuss if there is a good reason for us to have this feature and if so, raise an issue with `torch-spyre`.

spyre_inference/v1/attention/backends/spyre_attn.py
bohnstingl: I think this was an automatic lint fix. Let me double-check.

tdoublep (approved):

bohnstingl:
@joerunde, I really don't understand how the CI works. I played around with this PR today for some time.
Your commit `c8762e14ef2f150c575a14c50fa5cd8c84270fb3` passed the tests, then I tried to reduce some overhead code with alternative approaches, all failed. After that, I reverted everything back and ended up with commit `1a793b5bc33f049af5c45b529c7bec2ddb059724`, which is literally identical to your commit, except some commented lines.. 

```
diff --git a/spyre_inference/v1/worker/spyre_model_runner.py b/spyre_inference/v1/worker/spyre_model_runner.py
index 8007485..11c4509 100644
--- a/spyre_inference/v1/worker/spyre_model_runner.py
+++ b/spyre_inference/v1/worker/spyre_model_runner.py
@@ -180,14 +180,8 @@ class _SpyreModelWrapper:
         then convert the resulting logits back to CPU
         for downstream sampling.
         """
-        # if hidden_states is not None:
-        #     hidden_states = hidden_states.contiguous()
         hidden_states = convert(hidden_states, device=self._spyre_device)
-        # if hidden_states is not None:
-        #     hidden_states = hidden_states.contiguous()
         logits = self._model.compute_logits(hidden_states, *args, **kwargs)
-        # if logits is not None:
-        #     logits = logits.contiguous()
         return convert(logits, device="cpu")
```

However, the CI tests still fail for my commit. Do you have any magic insights here? This really freaks me out :sweat_smile:

PR #313, opened 2026-06-26
:sparkles: Enable facebook/opt-125m upstream test

joerunde:
## Description

Enables the `facebook/opt-125m` upstream model test by addressing three independent failures:

- **`OPTLearnedPositionalEmbedding.forward` patch** (`spyre_inference/models/opt.py`): OPT does `positions + self.offset` on int64; torch-spyre has no integer `aten.add` kernel, which aborts the SDSC scheduler. Do the add on CPU; the resulting indices feed the on-device embedding lookup. The genuinely OPT-specific piece.
- **CPU dispatcher for `vllm::unified_attention_with_output`** (`spyre_inference/custom_ops/attention_dispatch.py`): `SpyreQKVParallelLinear` D2Hs q/k/v (downstream `.split()` / scatter need CPU), so the opaque attention op dispatches to CPU — but the kernel is only registered for `PrivateUse1` and boot fails. New CPU impl moves inputs back to Spyre, re-dispatches, and mirrors the mutated outputs back to the caller's CPU buffers. Not OPT-specific — fixes the same boot failure for any model whose QKV linear D2Hs.
- **`SpyreUnquantizedEmbeddingMethod`** (`spyre_inference/custom_ops/vocab_parallel_embedding.py`): For `tie_word_embeddings=True` (OPT, GPT-2), `lm_head = embed_tokens` and `lm_head.quant_method.apply(...)` lands x (CPU, from `_SpyreModelWrapper`) and weight (Spyre) with mixed devices. New `apply` does the device conversion before/after the linear.

The latter two were originally drafted in #294. Cherry-picked here to keep this PR scoped to "enable opt-125m"; #294 will be closed and the remaining pieces landed separately.

## Related Issues

Relates to #294.

## Test Plan

```
uv run pytest -m "upstream and not distributed" 'tests/models/language/generation/test_common.py::test_models[False-False-5-2-facebook/opt-125m]'
```

- [x] `opt-125m` upstream test passes
- [x] `llama-3.2-1B` upstream test still passes (no regression)
- [x] `bash format.sh` clean (ruff, ty, typos, markdownlint, actionlint)

## Checklist

### DIFF
--- a/spyre_inference/__init__.py
+++ b/spyre_inference/__init__.py
@@ -37,10 +37,12 @@ def register():
 
 
 def register_ops():
-    """Register OOT custom ops for Spyre."""
-    from spyre_inference.custom_ops import register_all
+    """Register OOT custom ops and model patches for Spyre."""
+    from spyre_inference.custom_ops import register_all as register_custom_ops
+    from spyre_inference.models import register_all as register_model_patches
 
-    register_all()
+    register_custom_ops()
+    register_model_patches()
 
 
 def _init_logging():
--- a/spyre_inference/custom_ops/attention_dispatch.py
+++ b/spyre_inference/custom_ops/attention_dispatch.py
@@ -1,0 +1,85 @@
+# Copyright 2026 The Spyre-Inference Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


bohnstingl:
@joerunde I was playing with this model of this PR a bit and also some major rework of the CustomOps that we currently use. I figure that when using https://github.com/bohnstingl/spyre-inference/pull/2, the model runs for me without crashes (with enforce_eager=True). Could you maybe check that as well? If true, some of the changes here might not be required.

bohnstingl:
@joerunde The [CustomOp cleanup PR](https://github.com/torch-spyre/spyre-inference/pull/310) and the [workaround PR](https://github.com/torch-spyre/spyre-inference/pull/338) have landed. Can you please rebase and try again?

PR #314, opened 2026-06-26
:sparkles: Enable (longer context) granite support

joerunde:
## Description

Enables the upstream granite tests against `ibm-ai-platform/micro-g3.3-8b-instruct-1b` (single-card `test_granite.py::test_models` and TP=2 `test_basic_correctness.py::test_models_distributed`) by fixing a `RecursionError` in the attention backend that fired on longer prefills / larger batches. Also brings forward the platform `device_count` override and a small testing-plugin enhancement that were needed to make those tests actually run.

## Related Issues

`torch-spyre#220` / `torch-spyre#1371-3` — symbolic-offset support for `spyre.overwrite`; would let us revert to an on-device scatter and remove the CPU staging buffer added here.

## Changes

**Rewrite attention scatter-write** (`spyre_inference/v1/attention/backends/spyre_attn.py`)
- Per-token scatter into `output` had no working primitive on Spyre dim=0: `output[i:j] = ...` / `narrow().copy_(...)` silently write everything to row 0, and `torch.ops.spyre.overwrite` is deprecated. Its `compile_once` wrapper compiles one SDSC binary per unique `(shape, offset)` under `specialize_int=True`; once vLLM's model compile has filled the dynamo cache the kernel falls back to the same `compile_once` wrapper and recurses to Python's stack limit. Granite hit this in routine prefill; llama happened to compile-cache around it.
- `_online_softmax_attention` now builds the full per-attention-layer result in a CPU `output_cpu` tensor and copies it back to the Spyre output buffer in a single `output.copy_(...)` at the end of the per-sequence loop.
- Considered (and rejected) bumping `accumulated_cache_size_limit` to keep the on-device scatter: it works for short tests but compiles `N` binaries for a `query_len=N` prefill, which is unworkable as the on-device KV cache cap is raised (128K context = 128K binaries). When `torch-spyre` lands symbolic-offset overwrite, revisit and move the final write back on-device.

**Platform `device_count()` override** (`spyre_inference/platform.py`)
- `CpuPlatform.device_count()` returns 1 (CPU = single device); upstream gates like `@multi_gpu_test(num_gpus=2)` skipped on multi-card hosts as a result. Override to return `torch.spyre.device_count()` so TP-aware upstream tests actually run.

**Plugin support for tuple-parametrize override** (`tests/plugin/spyre_testing_plugin/pytest_plugin.py`)
- Some upstream tests use a single composite parametrize key like `@pytest.mark.parametrize("model, distributed_executor_backend, attention_backend, test_suite, extra_env", [...])`. The plugin's `override` hook only matched single-name keys, so the YAML couldn't redirect those tests at our model. Extends the matcher to recognise comma-joined keys (normalised on both sides) and to emit tuple values when the key is composite.

**Enable granite upstream tests** (`tests/plugin/spyre_testing_plugin/upstream_tests.yaml`)
- `test_granite.py::test_models`: `mandatory_pass` on `micro-g3.3-8b-instruct-1b` at fp16, `max_tokens=2`. Was a `model: []` placeholder waiting on the recursion fix.
- `test_basic_correctness.py::test_models_distributed`: `mandatory_pass` on `micro-g3.3-8b-instruct-1b` × `mp` executor × `CUSTOM` attention backend × test_suite `L4`, with `enable_prompt_embeds=[true]` skipped (prompt embeddings not yet supported on Spyre).
- `test_basic_correctness.py::test_models` remains skipped — its hardcoded ~5042-token prompt does not fit `MAX_MODEL_LEN_CAP=128`; orthogonal to this PR.
- No GHA workflow change needed: the existing `Upstream distributed vLLM tests` matrix entry (`-m "upstream and distributed"` on `spyre_pf_x2__pvc_1tb_x1`) already covers the newly-enabled TP=2 test, and `Upstream model vLLM tests` covers the granite single-card test.

**Doc correction** (`CLAUDE.md`)
- The "Spyre-Specific Constraints" list claimed "No tensor parallelism: Custom linear layers assume TP=1", which contradicts the in-tree `SpyreCommunicator` / `all_reduce` implementation documented at `docs/architecture/index.md` "Distributed (TP)". Corrected to "TP≥1 supported; DP>1 rejected in `TorchSpyrePlatform.check_and_update_config`".

## Test Plan

Local verification on a 2-Spyre host:

- [x] `uv run pytest -m "upstream and granite"` → 1 passed, 2172 deselected in 113s
- [x] `uv run pytest -m "upstream and distributed and model"` → 1 passed, 1 skipped (prompt-embeds), 2171 deselected in 129s
- [x] `uv run pytest -m "not upstream" tests/test_spyre_attn.py` → 73 passed, 24 skipped in 503s (no regression on the attention matrix from the CPU-buffer rewrite)
- [x] `bash format.sh` → ruff, ruff format, typos, markdownlint, GHA workflow lint, ty check all pass

## Checklist

- [X] I have added tests for my changes (if applicable) — covered by the newly-enabled upstream tests
- [X] I have updated the documentation (if applicable) — CLAUDE.md TP claim corrected

### DIFF
--- a/spyre_inference/v1/attention/backends/spyre_attn.py
+++ b/spyre_inference/v1/attention/backends/spyre_attn.py
@@ -777,9 +777,7 @@ class SpyreAttentionImpl(AttentionImpl[SpyreAttentionMetadata]):
         # result on CPU and bulk-copy at the end of the per-sequence loop.
         # Revisit when torch-spyre lands symbolic-offset overwrite
         # (torch-spyre#220 / #1371-3).
-        output_cpu = torch.zeros(
-            num_actual_tokens, num_heads, head_size, dtype=output.dtype, device="cpu"
-        )
+        output_cpu = torch.zeros_like(output, device="cpu")
 
         for seq_idx in range(num_seqs):
             # Most-naive implementation: no parallelization
--- a/spyre_inference/v1/attention/backends/spyre_attn.py
+++ b/spyre_inference/v1/attention/backends/spyre_attn.py
@@ -767,25 +767,16 @@ class SpyreAttentionImpl(AttentionImpl[SpyreAttentionMetadata]):
             "attention_mask_tiles must be precomputed by the metadata builder"
         )
 
-        # Per-token scatter into a spyre-resident `output` is unworkable today:
-        #   - `output[i:j] = ...` / `narrow(...).copy_(...)` silently smear the
-        #     source across the whole buffer on dim=0 (writes hit row 0).
-        #   - `torch.ops.spyre.overwrite(..., dims=[0], offsets=[q_start+i])`
-        #     does write to the right row, but its compile_once wrapper traces
-        #     once per unique (shape, offset) under `specialize_int=True`. Each
-        #     trace consumes a slot in dynamo's accumulated_cache_size_limit
-        #     (default 256, untouched by torch_spyre._autoload's per-fn 1024
-        #     bump). vLLM's model compile already fills most of that budget;
-        #     these calls tip it over, dynamo bails out to the eager kernel —
-        #     which is the same compile_once wrapper — and we recurse until
-        #     Python's recursion limit (granite repro). Bumping the limit
-        #     unblocks short tests but doesn't scale: with chunked prefill
-        #     disabled, query_len reaches the full prompt length, so a 128K
-        #     context would compile 128K SDSC binaries (hours of compile +
-        #     several GB of cache). Build the full result on CPU and write
-        #     `output` in one bulk copy instead. Revisit when torch-spyre
-        #     lands symbolic-offset overwrite (issues #220 / #1371-3) — one
-        #     compiled binary for any offset would let us scatter on-device.
+        # Scattering into `output` on Spyre dim=0 has no working primitive:
+        # `output[i:j] = ...` and `narrow().copy_(...)` silently write to row 0;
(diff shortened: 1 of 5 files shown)


spyre_inference/v1/attention/backends/spyre_attn.py
```diff
@@ -766,6 +767,20 @@ def _online_softmax_attention(
...
+        # compiles one SDSC binary per unique offset, which recurses past the
+        # dynamo cache limit once vLLM's model compile has filled it. Raising
+        # the limit unblocks short tests but compiles N binaries for a
+        # query_len=N prefill, which doesn't scale to long contexts. Stage the
+        # result on CPU and bulk-copy at the end of the per-sequence loop.
+        # Revisit when torch-spyre lands symbolic-offset overwrite
+        # (torch-spyre#220 / #1371-3).
+        output_cpu = torch.zeros(
+            num_actual_tokens, num_heads, head_size, dtype=output.dtype, device="cpu"
+        )
```
bringlein:
```suggestion
        output_cpu = torch.zeros_like(output, device="cpu")
```

Just small comment: This version may be more robust to layout changes? Otherwise, I would not expect that `output` shape changes.

bringlein (approved):
LGTM, thanks for digging into the recursion of  `_overwrite`. 

Yes, `torch.ops.spyre.overwrite` is deprecated, but we had the hope using it saves us unnecessary device-host-device cycles. However, since slicing is not supported yet on the device, they are necessary anyhow for the time being.

PR #318, opened 2026-06-29
:construction: Swap CI image to torch-spyre, bump past breaking changes

joerunde:
## Description

This PR bumps torch-spyre to the latest main, but it requires updating the base image to pick up newer runtime libraries in order to compile successfully.

~~(TBD working on getting a valid set of runtime libraries- new runtime regression introduced by deeptools needs at least commit a438898 to function)~~

Working library versions:
```
ibm-aiu-toolbox-e2e-2.0.0-0.main.1+25.2bcc25f_0.el10
ibm-deeptools-2.0.0-0.main.1+1245.0a9da5e_0.el10
ibm-deeptools-devel-2.0.0-0.main.1+1245.0a9da5e_0.el10
ibm-flex-2.0.0-0.main.1+349.8a6d706_0.el10
ibm-flex-devel-2.0.0-0.main.1+349.8a6d706_0.el10
ibm-libaiupti-2.0.0-0.main.1+12.a3f540d_0.el10
ibm-senlib-core-2.0.0-0.main.1+178.ca4666a_0.el10
ibm-senlib-dd2-2.0.0-0.main.1+178.ca4666a_0.el10
ibm-senlib-headers-2.0.0-0.main.1+178.ca4666a_0.el10
ibm-spyre-comms-1.0.0-0.main.1+62.286e381_0.el10
ibm-spyre-comms-devel-1.0.0-0.main.1+62.286e381_0.el10
ibm-spyre-comms-test-1.0.0-0.main.1+62.286e381_0.el10
```

## Checklist

- [ ] I have read the [contributing guidelines](https://torch-spyre.github.io/spyre-inference/contributing/)
- [ ] My code follows the project's code style (run `bash format.sh`)
- [ ] I have added tests for my changes (if applicable)
- [ ] I have updated the documentation (if applicable)
- [ ] My commits include a `Signed-off-by:` line (DCO compliance)

PR #320, opened 2026-06-30
ci-cd: avoid circular dependency by not using the spyre-inference image to build spyre-inference

HarikrishnanBalagopal:
## Description

We should not use `spyre-inference` image to build `spyre-inference` because breaking changes in deeptools, flex, torch-spyre that require an corresponding fix in spyre-inference can't be tested unless we merge the PR into spyre-inference. And for merging the PR we need passing tests, which requires a proper runner image, which require image builds to pass, which requires spyre-inference build to pass. This creates a chicken and egg problem

## Related Issues

Related change in torch-spyre repo https://github.com/torch-spyre/torch-spyre/pull/2794

## Checklist


joerunde:
cherry-picked and merged into #318

